PR #4, opened 2022-05-02
Add sphinx quickstart template

grst:


ilia-kats:
Does it make sense to have a `references.bib` with Scanpy if it's not used in the template?

grst:
it's just meant as an example

PR #5, opened 2022-05-02
Add simple API scaffold with pp, tl, and pl

gtca:
(An updated version of) the proposed code scaffold:

```
package_name
    ├- pp (preprocessing)
    │   └- basic_plot.py
    |
    ├- tl (for tools)
    │   └- basic_tool.py
    |
    └- pl (for plotting)
        └- basic_plot.py
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+root = true
+
+[*.py]
+indent_style = space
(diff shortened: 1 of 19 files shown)


ilia-kats:
should we also annotate return types? Just to encourage this practice?

PR #9, opened 2022-05-02
add pre-commit checks and checks summary

giovp:


grst:
Come have some beers!

{{cookiecutter.project_name}}/.pre-commit-config.yaml
```diff
@@ -1,23 +1,82 @@
+fail_fast: false
+default_language_version:
+  python: python3
+default_stages:
+  - commit
+  - push
+minimum_pre_commit_version: 2.9.3
```
ivirshup, line 7: Update this?

@michalk8 says it's at 2.16 at least

```suggestion
minimum_pre_commit_version: 2.16.0
```

{{cookiecutter.project_name}}/.pre-commit-config.yaml
```diff
@@ -1,23 +1,82 @@
...
+    hooks:
+      - id: yesqa
+        additional_dependencies:
+          [
+            flake8-tidy-imports,
+            flake8-docstrings,
+            flake8-rst-docstrings,
+            flake8-comprehensions,
+            flake8-bugbear,
+            flake8-blind-except,
```
ivirshup, line 41: Can these go in the .flake8 file?

ivirshup: If we add them there, do these have to be dev dependencies? Probably don't want that.

giovp: they can't go in `.flake8` apparently, only local plugin.

flying-sheep (approved):
I think we should merge this before more conflicts happen

giovp:
yeah we should wait for #30 and then merge that

giovp:
so pre-commit are failing locally, I'll have a look later

giovp:
I was still working on it 😅

PR #15, opened 2022-05-03
Update README template

grst:
Suggestion for a basic README with some essential information, linking to the docs where appropriate.


flying-sheep:
looks like a problem with conflicting pre-commit checks here

PR #36, opened 2022-05-03
add pre-commit check to CI

giovp:


flying-sheep:
Did you want to add this to the template? because this repo is already using pre-commit.ci:
grafik

Issue #28, opened 2022-05-03
Switch to src/ layout


flying-sheep:
See here for places recommending it:

- PyPA [discussion](https://github.com/pypa/packaging.python.org/issues/320#issuecomment-495990983), [protocol](https://docs.google.com/document/d/1Wz2-ECkicJgAmQDxMFivWmU2ZunKvPZ2UfQ59zDGj7g/edit#heading=h.2cgqnlxl8y3e)

  (Note that “single module” uses official terminology in which “module” means `x.is_file() and x.suffix == '.py'` and  “package” means `x.is_dir() and x.join('__init__.py').is_file()`.)

- [Ionel’s blog](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure)
- [Hypermodern Python](https://cjolowicz.github.io/posts/hypermodern-python-01-setup/#creating-a-package-in-src-layout) defers to [Hynek’s blog](https://hynek.me/articles/testing-packaging/)

Furthermore I think:

- The symmetry makes sense: `ls *(/)` will yield `src` `tests` `docs`, package specific names start appearing inside of those
- adding `{project}/src` to PYTHONPATH or doing `pip install -e .` is safer than adding `{project}`

(Resolution: Try it out after all PRs touching the code got merged)

flying-sheep:
Since finding the discussion, I also found that both the Hypermodern Python package guide and Ionel’s blog (i.e. all places that I found that gives reasons for choosing a layout) picked this one.

ivirshup:
It may not have been clear due to the hybrid meeting (which we definitely didn't prepare well for), but I think everyone else was against this. I think I'm going to revert this, though we can always add  it if it gets some support.

flying-sheep:
You said “let’s try it”.

ivirshup:
Ah yeah, so what I had meant there (and I think was in response to others) was: "lets try it out for a project before making it a part of the template".

flying-sheep:
Aah, I see! How far did you get with the stats package yesterday?

flying-sheep:
The more I read about it the more convinced I become: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#src-layout

grst:
> It seems unnecessary and reminded me of Java

That was exactly my feeling, but Hynek makes a few good points. 

I think I'm slightly in favor of `src` now, but the question remains whether it's worth the effort changing something.

flying-sheep:
[...]

PR #39, opened 2022-05-04
Revert "Switch to src/ layout"

ivirshup:
Reverts scverse/cookiecutter-scverse#37


ivirshup:
See: https://github.com/scverse/cookiecutter-scverse/issues/28#issuecomment-1117154942

PR #58, opened 2022-05-17
add main as default for docs

giovp:


{{cookiecutter.project_name}}/tests/test_basic.py
```diff
@@ -1,9 +1,12 @@
...
 
 
 def test_package_has_version():
     {{cookiecutter.package_name}}.__version__
 
 
-def test_fails():
-    assert 1 == 0
+@pytest.mark.skip(reason="This decorator should be removed when test passes.")
+def test_example():
```
Zethson, line 11: I still don't see any reason why we would add a skipped/failing test when a single test the passes is good enough. Like people won't magically look for the reason as to why a test has been skipped. Either people want to write tests or they don't.

To me this is noise. @giovp your skip solution is of course the one that I prefer 100000% compared to the failing test.

I'd just have a single passing test.

Zethson (approved):
LGTM. Could also recommend global organization/username secrets instead of repository secrets. But that's a minor detail.

PR #61, opened 2022-05-23
Remove macOS from CI

ivirshup:
Removing macOS from github actions, since it's quite expensive for private repos.


grst:
Wasn't aware that this also counts for the free minutes, but macOS runners indeed consume 10 minutes per actual minute :scream:

PR #62, opened 2022-05-23
Add issue templates

### LINKED ISSUE #57: Create issue templates
I think this repo should provide sensible default issue templates and link to the scverse discourse for help requests.

grst:
Close #57


{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,71 @@
...
+               * Version of AnnData (`anndata.__version__`)
+               * Python version (*e.g. 3.7, 3.8*)
+               * OS (*e.g. Arch Linux, Windows 10, MacOS*)
+          placeholder: |
+              {{ cookiecutter.package_name }}: 0.7.0
+              numpy: 1.22.3
+              pandas: 1.4.2
+              AnnData: 0.7.6
+              Python: 3.7
+              OS: Arch Linux
```
grst, line 71: @Zethson proposed of asking to use `session_info` here. The question is then if `session_info` should become a requirement in `pyproject.toml`

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,71 @@
...
+      id: repex
+      attributes:
+          label: Minimal reproducible example
+          description: A code snipped that we can copy&paste to reproduce the bug without having any data
+          render: python
+          placeholder: |
+              import scanpy as sc
+              adata = sc.datasets.pbmc3k()
+
+              sc.do_something_that_doesnt_work(adata)
```
grst, line 33: I was first tempted to do 
```
import {{ cookiecutter.package_name }}

{{ cookiecutter.package_name }}.do_something_that_doesn't work(adata)
```

but not every package provides example datasets.

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,71 @@
...
+body:
+    - type: markdown
+      attributes:
+          value: |
+              **Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports)
+              detailing how to provide the necessary information for us to reproduce your bug. In brief:
+                * Please provide exact steps how to reproduce the bug in a clean Python environment.
+                * In case it's not clear what's causing this bug, please provide the data or the data generation procecure.
+                * Sometimes it is not possible to share the data but usually it is possible to replicate problems on publicly
+                  available datasets or to share a subset of your data.
```
grst, line 13: adapted from mudata

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
Zethson: Also not sure. 
1. session_info is not a dependency that is required to use the package
2. It is easier for newbies if they only install the package and then can use session_info instead of awkwardly having to install this as well.

I would lean towards making it a dependency. What do the others think?

grst: In principle, I agree with making it a dependency, but there's also https://github.com/scverse/cookiecutter-scverse/issues/57#issuecomment-1136161434

@scverse/core-devs ?

grst:
I think we agreed to stick with the simple, explicit version placeholder for now. 
Once we have #72, we can change this.

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
ivirshup: I'd just make it a dependency for now.

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,71 @@
...
+          placeholder: |
+              ---------------------------------------------------------------------------
+              AssertionError                            Traceback (most recent call last)
+              <ipython-input-5-78750baef17a> in <module>
+              ----> 1 adata = sc.do_something_that_doesnt_work(adata)
+
+              /opt/conda/lib/python3.8/site-packages/scanpy/_preprocessing.py in do_something_that_doesnt_work(adata)
+                  438
+              --> 439     raise AssertionError("Demo function that fails")
+                  440
```
ivirshup, line 50: I'm not sure this is a lot better than just having free-form markdown.

While many bug reports fit "ran this code, got this error" not all do. Especially difficult to solve ones. How does "incorrect thing works" or "this works, this works, this doesn't" fit into this template?

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+    - name: Scverse Community Forum
+      url: https://discourse.scverse.org/
+      about: If you have questions about “How to do X”, please ask them here.
```
ivirshup, line 5: I think this needs a little more consideration. What we're currently doing for discourse is creating tags for each package. That's a manual process, and one that probably shouldn't be automate on package instantiation.

See current organization for discourse in this comment and the linked google doc: https://github.com/scverse/governance/issues/26#issuecomment-1150123357

tbh, I think using issues for discussion is probably fine for the first little while.

ivirshup (commented):
Looks mostly good. 

Had two comments re: structure of bug report and use of discourse.

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/config.yml
grst: It would be a nice way to advertise discourse as *the* place for single-cell analysis discussions, though. 
We could also have a tag for "other package" and start creating dedicated tags once the traffic increases.

ivirshup: @adamgayoso, what do you think?

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
grst: All fields except for the first one are optional and you can write free-form markdown there.

grst:
Template can be previewed here:
https://github.com/scverse/cookiecutter-scverse-instance/blob/8d982c9adc9587fcd37cee10cf0dbf242cb99710/.github/ISSUE_TEMPLATE/bug_report.yml

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
ivirshup: My preference here would be:

One required field with description of bug. This includes a template that looks like:

`````
## Description

## Example

```python
import {pkg}

# Minimal example to reproduce bug
````

```pytb
# Traceback from running code
```
`````

And a required field for version info.

To me, Text, Code block, Error block may be the most common bug report format, but I'm not sure it's over 50% prevalence. I think putting too much structure here makes it harder for people to report, since they'll try to fit their problem into structure.

File: {{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/config.yml (hunk 5 above)
Comment (adamgayoso) [reply {{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/config.yml:5]: yes everyone should use it

File: {{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml [hunk 6]
```diff
@@ -0,0 +1,118 @@
...
+              adata = sc.datasets.pbmc3k()
+
+              sc.do_something_that_doesnt_work(adata)
+
+    - type: textarea
+      id: stacktrace
+      attributes:
+          label: The error message produced by the code above
+          description: Please paste the entire error message
+          render: pytb
```
giovp, line 40: this is flagged as unknown in my vscode (I have github extension but still maybe not recognized)

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,118 @@
...
+      attributes:
+          label: Version information
+          description: |
+              Please paste below the output of
+
+              ```python
+              import session_info
+              session_info.show(html=False, dependencies=True)
+              ```
+          placeholder: |
```
giovp, line 63: would shorten this examle by a lot (like adding ... after 3 packages)

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/config.yml
giovp: I agree and tbh I would be very strict to what can be reported as github issue: only bug report and feature request, everything else should be discourse IMHO. How do we do that unclear, as in we can't not answer issues.

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -19,7 +19,11 @@ maintainers = [
 urls.Documentation = "https://{{ cookiecutter.project_name }}.readthedocs.io/"
 urls.Source = "{{ cookiecutter.project_repo }}"
 urls.Home-page = "{{ cookiecutter.project_repo }}"
-dependencies = ["anndata"]
+dependencies = [
+    "anndata",
+    # for debug logging (referenced from the issue template)
```
giovp, line 24:
```suggestion
```

giovp (approved):
minor nitpick, feel free to disregard, I'd get this merged asap.

grst:
By now, I tend to side with @ivirshup: Rather have a less-structured bug report form - I think it's still better to receive low quality bug reports than not receiving any. 

But a markdown template within an issue form is also ugly. Maybe just make it very explicit in the forms description which fields are optional?

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
grst: it renders fine: https://github.com/scverse/cookiecutter-scverse-instance/blob/8d183cec232d72485a82d18cef72201ceff0e23d/.github/ISSUE_TEMPLATE/bug_report.yml

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,118 @@
...
+              ---------------------------------------------------------------------------
+              AssertionError                            Traceback (most recent call last)
+              <ipython-input-5-78750baef17a> in <module>
+              ----> 1 adata = sc.do_something_that_doesnt_work(adata)
+
+              /opt/conda/lib/python3.8/site-packages/scanpy/_preprocessing.py in do_something_that_doesnt_work(adata)
+                  438
+              --> 439     raise AssertionError("Demo function that fails")
+                  440
+
```
ivirshup, line 51:
```suggestion
name: Bug report
description: Report something that is broken or incorrect
labels: bug
body:
    - type: markdown
      attributes:
          value: |
              **Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports)
              detailing how to provide the necessary information for us to reproduce your bug. In brief:
                * Please provide exact steps how to reproduce the bug in a clean Python environment.
                * In case it's not clear what's causing this bug, please provide the data or the data generation procecure.
                * Sometimes it is not possible to share the data but usually it is possible to replicate problems on publicly
                  available datasets or to share a subset of your data.

    - type: textarea
      id: report
      attributes:
          label: Report
          description: A clear and concise description of what the bug is.
      validations:
          required: true

```

@grst, would cutting it down to this be enough? Could include a bit more.

grst:
Ok, let's go with you suggestion, @ivirshup. 

We can always extend the template later if it turns out to be necessary.

PR #69, opened 2022-05-30
Separate building from testing

ivirshup:


{{cookiecutter.project_name}}/.github/workflows/build.yaml
```diff
@@ -0,0 +1,23 @@
...
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - name: Set up Python 3.10
+              uses: actions/setup-python@v2
+              with:
+                  python-version: "3.10"
+            - name: Install build dependencies
+              run: python -m pip install --upgrade pip wheel twine build
+            - name: Build package
```
Zethson, line 20: Can we cache this?

ivirshup: Probably

PR #81, opened 2022-07-05
Update developer documentation

grst:
Close #24 
Close #17
Close #51

PR #87, opened 2022-07-08
Remove autopep8

grst:
Black already covers most (if not all) of what autopep does. 
Plus in weird cases the tools can be conflicting.


grst:
Removed from the list of pre-commit checks in #81

PR #90, opened 2022-07-20
Correct git version requirement in README.md

marcovarrone:
In the documentation of README, I imagine that the version requirement for git is because of the "--initial-branch" parameter introduced in version 2.28. 
However, the documentation says git >= 2.38, a version of git that doesn't exist yet.


grst:
Thanks!

PR #97, opened 2022-10-07
.flake8 changes

flying-sheep:
Changes

- grouped more things together that conflict with black
- The comment “i, j, k, l is reasonable indexing notation” was misleading, E741 doesn’t actually prevent variables from being named, i, j, or k
- F401 (unused import) is a good rule, helps getting rid of unused dependencies. One should use `__all__` in public modules
- B008 (no function calls in argument defaults) is also a good rule. One can use `extend-immutable-calls` to extend the list of allowable calls.

### DIFF
--- a/{{cookiecutter.project_name}}/.flake8
+++ b/{{cookiecutter.project_name}}/.flake8
@@ -41,7 +41,7 @@ per-file-ignores =
     tests/*: D
     */__init__.py: F401
 extend-immutable-calls =
-    # If you see B008 but know 
+    # Add functions returning immutable values here to avoid B008
     pathlib.Path
     Path
 rst-roles =
--- a/{{cookiecutter.project_name}}/.flake8
+++ b/{{cookiecutter.project_name}}/.flake8
@@ -10,32 +10,28 @@ ignore =
     E501
     # whitespace before : -> black does not adhere to PEP8
     E203
+    # line break before binary operator -> black does not adhere to PEP8
+    W503
     # missing whitespace after ,', ';', or ':' -> black does not adhere to PEP8
     E231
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_name}}/.flake8
```diff
@@ -10,32 +10,28 @@ ignore =
...
     E203
+    # line break before binary operator -> black does not adhere to PEP8
+    W503
     # missing whitespace after ,', ';', or ':' -> black does not adhere to PEP8
     E231
     # continuation line over-indented for hanging indent -> black does not adhere to PEP8
     E126
-    # E266 too many leading '#' for block comment -> this is fine for indicating sections
+    # whitespace before ':' -> black does not adhere to PEP8
+    E203
```
grst, line 20:
```suggestion
```

this one is duplicate (L10)

{{cookiecutter.project_name}}/.flake8
```diff
@@ -10,32 +10,28 @@ ignore =
...
     E262
     # Do not assign a lambda expression, use a def -> lambda expression assignments are convenient
     E731
-    # allow I, O, l as variable names -> I is the identity matrix, i, j, k, l is reasonable indexing notation
+    # allow I, O, l as variable names -> I is the identity matrix
     E741
     # Missing docstring in public package
     D104
-    # ... imported but unused
-    F401
```
grst, line 26: isort takes care of this... any chance this is required to make this compatible? E.g. if flake8 runs before isort in pre-commit.

grst: autoflake does. But flake8 runs after it, so there shouldn't be any issue.

flying-sheep (commented):
…

PR #102, opened 2022-10-10
Allow user to specify a package name that doesn’t match the project name

flying-sheep:
E.g. if you specify `project_name: mycompany-foo` and `package_name: foo`, this should make it work


ivirshup:
@adamgayoso, you were just modifying this right?

grst (approved):

PR #110, opened 2022-10-20
Prefer f-literals to +-concatenation

flying-sheep:
Using `+` to concatenate strings is a bad habit: It breaks when you use a variable that isn’t an instance of `str`, and is more noisy than using f literals.


Zethson (approved):
It's weird that f-strings weren't used in the first place here. Almost makes me wonder that there must've been some reason to not use them^^

But I don't see why not. Let's merge and we can revert if this breaks something.

Issue #119, opened 2022-11-01
Pytest doesn't find project's module


marcovarrone:
After a fresh setup of the project using cookiecutter, I run the `pytest` command and at test_basic.py I get an error: 
`ModuleNotFoundError: No module named ... `

Basically, the environment from pytest doesn't have the project's location in the system path and fails to find the path to the code.

grst:
I think you should install the package with `pip install -e` before testing. Maybe we need to improve docs on that. 

Any other opinions @flying-sheep @ivirshup?

ivirshup:
@grst I think the docs could definitely be clearer here.

PR #120, opened 2022-11-01
Fix pytest finding project's module

marcovarrone:
Fixes [#119](https://github.com/scverse/cookiecutter-scverse/issues/119).

### DIFF
--- a/{{cookiecutter.project_name}}/pyproject.toml
+++ b/{{cookiecutter.project_name}}/pyproject.toml
@@ -65,7 +65,7 @@ addopts = [
     "--import-mode=importlib",  # allow using test files with same name
 ]
 pythonpath = [
-    "src"
+    ".", "src"
 ]
 
 [tool.isort]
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -64,6 +64,9 @@ xfail_strict = true
 addopts = [
     "--import-mode=importlib",  # allow using test files with same name
 ]
+pythonpath = [
+    ".", "src"
```
flying-sheep, line 68: The whole reason for the `src` layout is so that `.` doesn’t have to be in PYTHONPATH under any circumstances: https://docs.pytest.org/en/latest/explanation/goodpractices.html

```suggestion
    "src"
```

flying-sheep (changes requested):

{{cookiecutter.project_name}}/pyproject.toml
grst: yes, but do we need  `pythonpath` at all, if we say everyone should install the package? (See discussion #119).

marcovarrone: I agree, feel free to close it :)

PR #131, opened 2022-12-09
Add post gen hook with detailed instructions

### LINKED ISSUE #83: Add post-gen hook that prints further instructions
After creating the cookiecutter template, the developer needs to connect a bunch of online services (readthedocs, codecov, and potentiall pre-commit.ci (#50)). 

While this is listed in the README, I think it would be nice to remind the user after running the cookiecutter CLI. 

This can easily be achieved using a post-gen hook: 
https://cookiecutter.readthedocs.io/en/1.7.2/advanced/hooks.html

grst:
Close #83


giovp (approved):
ok for me but don't see as super problematic to have rich as dependency for the template

grst:
it's more about people forgetting to install it (needs to be done manually along cruft) and then the whole creation fails at the very end because of that.

PR #137, opened 2023-01-09
Fix template sync due to lacking permissions

grst:
In a test-run, the template sync failed since apparently
the `sync.yaml` action is not permitted to make changes to files in the
`.github/workflows` directory:
https://github.com/icbi-lab/infercnvpy/actions/runs/3875431171/jobs/6607954535

This is a bit of a problem for which we still need to figure out a
solution. One of them is to use a personal access token. Another one
could be to centralize template sync, which I would have liked to avoid.

For now, I just rolled back the changes to the workflow files since
the last release since they were mostly of cosmetic nature.

### DIFF
--- a/{{cookiecutter.project_name}}/.github/workflows/sync.yaml
+++ b/{{cookiecutter.project_name}}/.github/workflows/sync.yaml
@@ -43,4 +43,4 @@ jobs:
                       manually merge these changes.**
 
                       For more information about the template sync, please refer to the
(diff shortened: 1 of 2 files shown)


grst:
@Zethson, another one please.

Actually the underlying issue is pretty nasty -- something I wasn't aware of when setting up template sync. But this should at least be enough to make the nightly cron job pass again for now.

Zethson (approved):
All right.

grst:
thx

PR #144, opened 2023-01-17
Pin older node version for prettier pre-commit hook

### LINKED ISSUE #143: BUG: pre-commit/repoxoi92sgz/node_env-default/bin/node: libc.so.6: version `GLIBC_2.28' not found
I fail to initialise a repo from this template because pre-commit fails on `git commit ...`. 

```bash
[INFO] Installing environment for https://github.com/pre-commit/mirrors-prettier.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('~/.cache/pre-commit/repoxoi92sgz/node_env-default/bin/node', '~/.cache/pre-commit/repoxoi92sgz/node_env-default/bin/npm', 'install', '--dev', '--prod', '--ignore-prepublish', '--no-progress', '--no-save')
return code: 1
stdout: (none)
stderr:
    ~/.cache/pre-commit/repoxoi92sgz/node_env-default/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by ~/.cache/pre-commit/repoxoi92sgz/node_env-default/bin/node)

Check the log at ~/.cache/pre-commit/pre-commit.log
```

The solution presented here https://github.com/jupyterlab/jupyterlab/issues/12675  is adding
```
default_language_version:
  node: system
```
to the top of `.pre-commit-config.yaml`.

However, that solution also fails with:

```
[INFO] Installing environment for https://github.com/pre-commit/mirrors-prettier.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('npm', 'install', '--dev', '--prod', '--ignore-prepublish', '--no-progress', '--no-save')
return code: 1
stdout:
    Executable `npm` not found
stderr: (none)
```

I tried to solve this by installing `conda install -c conda-forge nodejs` - however, this also fails:

```
prettier.................................................................Failed
- hook id: prettier
- exit code: 1

prettier requires at least version 14 of Node, please upgrade
```

Is `prettier` really necessary?

grst:
Fix #143

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,10 @@
-repos:
-    - repo: https://github.com/pre-commit/mirrors-prettier
-      rev: v3.0.0-alpha.4
+- repo: https://github.com/pre-commit/mirrors-prettier
+      rev: v2.5.1
       hooks:
           - id: prettier
+            # Newer versions of node don't work on systems that have an older version of GLIBC
+            # (in particular Ubuntu 18.04 and Centos 7)
+            # EOL of Centos 7 is in 2024-06, we can probably get rid of this then.
+            # See https://github.com/scverse/cookiecutter-scverse/issues/143 and
+            # https://github.com/jupyterlab/jupyterlab/issues/12675
+            language_version: "17.9.1"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
     - repo: https://github.com/pre-commit/mirrors-prettier
-      rev: v2.5.1
+      rev: v2.5.1  # Only update together with the pre-commit yaml in the tempalate! 
       hooks:
           - id: prettier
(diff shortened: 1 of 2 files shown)


Zethson (commented):
TBH I'm not sure whether I want the version pinned in the template. Maybe something that should rather be in the docs?
How many people are affected by this? I saw the issue...

grst:
It affects everyone on Ubuntu 18.04, which I'm afraid is still rather common.

Zethson:
Sadly you are correct.
The support for 18.x ends in April 2023 but is only EOL in 2028.

DW, I won't block this PR, but these kinds of fixes that at one point might get forgotten annoy me :) Thank you for looking into it!

grst:
18.10 should work though. 

Not sure about other OS. The linked jupyterhub issue also mentions amazon linux.
Maybe let's just float this around for a while and see if more people complain.

PR #154, opened 2023-03-02
Update docs version

adamgayoso:
[built docs](https://cookiecutter-scverse-instance--42.org.readthedocs.build/en/42/)

1. lower bound the theme version so that we get dark mode!
2. Use sphinx ext autodoc for typehint processing instead of sphinx-autodoc-typehints which we borrowed from scanpy at a time when sphinx didn't have this functionality. The third-party package also started processing numpydoctring return sections differently resulting in a colon in front of our custom freeform return
3. Edited the autosummary class template to remove headers for attr/meth names, which aren't necessary. Without this it looks like this on the right sidebar:

Screenshot 2023-03-02 at 10 55 02 AM

### DIFF
--- a/{{cookiecutter.project_name}}/docs/_templates/autosummary/class.rst
+++ b/{{cookiecutter.project_name}}/docs/_templates/autosummary/class.rst
@@ -39,9 +39,6 @@ Attributes
 
 {% for item in attributes %}
 
-{{ item }}
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
 .. autoattribute:: {{ [objname, item] | join(".") }}
 {%- endfor %}
 
@@ -55,9 +52,6 @@ Methods
 
 {% for item in methods %}
 {%- if item != '__init__' %}
-
-{{ item }}
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. automethod:: {{ [objname, item] | join(".") }}
 {%- endif -%}
--- a/{{cookiecutter.project_name}}/docs/conf.py
+++ b/{{cookiecutter.project_name}}/docs/conf.py
@@ -109,6 +109,7 @@ html_title = project_name
 html_theme_options = {
     "repository_url": repository_url,
     "use_repository_button": True,
+    "path_to_docs": "docs/",
 }
 
 pygments_style = "default"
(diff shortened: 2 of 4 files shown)


flying-sheep:
I don’t think `sphinx.ext.autodoc` is ready for prime time. I used it for a little package I maintain, and had to hack around its limitations in two instances. I think `sphinx-autodoc-typehints` (without `scanpydoc`’s by-now-unnecessary customization) just works:tm: and should be used for now.

E.g. I had to turn this wonderfully reasonable definition:

```py
from operator import add
from collections.abc import Iterable, Callable
from typing import TypeVar

A = TypeVar("A")
T = TypeVar("T")

class accumulate(...)
    def __init__(iterable: Iterable[A], func: Callable[[A, A], T] = add) -> None:
        ...
```

into this monstrosity to keep `sphinx.ext.autodocs` from choking:

```py
...

class _Adder:
    """This just exists to allow Sphinx to parse the function signature."""

    def __call__(self, a: A, b: A) -> T:
        a + b

    def __repr__(self) -> str:
        return "add"  # pragma: no cover

class accumulate(...)
    def __init__(iterable: Iterable[A], func: Callable[[A, A], T] = _Adder()) -> None:
        ...
```

adamgayoso:
@flying-sheep happy to keep sphinx autodoc typehints, but would you be able to update the typed returns extension on this repo?

This PR on sphinx autodoc typehints https://github.com/tox-dev/sphinx-autodoc-typehints/pull/311 causes an issue where : precedes all return descriptions.

flying-sheep:
I investigated further, and while using `'description'` helps with the problem I described, it doesn’t fix other problems like spurios “class not found” errors when using TypeVars. `sphinx-autodoc-typehints` is just more mature right now.

- https://github.com/sphinx-doc/sphinx/issues/10785
- https://github.com/sphinx-doc/sphinx/issues/10794

I’ll figure out the return problem.

Comparison:

<table
<thead><tr><th width="50%">

`sphinx-autodoc-typehints`

<th width="50%">

`autodoc_typehints = "description"`

<tbody><tr><td>

grafik

<td>

grafik

<tr><td>

*no log errors*

<td>

grafik

</table>

I think neither style is good. `sphinx-autodoc-typehints` is too visually busy, and `autodoc_typehints = "description"` breaks the otherwise consistent visual style for references to functions, classes, …

adamgayoso:
I didn't file an issue because the problem seems to be more specific to our custom typed returns extension, which is in this repo/scanpy/etc. You can see the "bug" in the left image above: `: Some integer value`

I can remove the change in this PR, but we should figure something out with the `:` issue.

flying-sheep:
It just seems to … be there. So we can remove it.

We should move that little customization to scanpydoc and test it properly, but for now, I’ll just fix it here.

flying-sheep (approved):

PR #162, opened 2023-04-13
Update how scverse citation is included in template

ivirshup:
Change how the scverse citation is added

Basically, I don't think it should be added to the readme of projects using the template, but do think it's reasonable to include it as the example citation. We could maybe make a suggestion to keep it, but wasn't sure how to word that.

### DIFF
--- a/{{cookiecutter.project_name}}/README.md
+++ b/{{cookiecutter.project_name}}/README.md
@@ -45,17 +45,6 @@ See the [changelog][changelog].
 For questions and help requests, you can reach out in the [scverse discourse][scverse-discourse].
 If you found a bug, please use the [issue tracker][issue-tracker].
 
-## Citation
-
-> t.b.a.
-
-You can cite the scverse publication as follows:
-
-> **The scverse project provides a computational ecosystem for single-cell omics data analysis**
->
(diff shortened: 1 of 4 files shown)


{{cookiecutter.project_name}}/README.md
```diff
@@ -45,17 +45,6 @@ See the [changelog][changelog].
...
-
-> t.b.a.
-
-You can cite the scverse publication as follows:
-
-> **The scverse project provides a computational ecosystem for single-cell omics data analysis**
->
-> Isaac Virshup, Danila Bredikhin, Lukas Heumos, Giovanni Palla, Gregor Sturm, Adam Gayoso, Ilia Kats, Mikaela Koutrouli, Scverse Community, Bonnie Berger, Dana Pe’er, Aviv Regev, Sarah A. Teichmann, Francesca Finotello, F. Alexander Wolf, Nir Yosef, Oliver Stegle & Fabian J. Theis
->
-> _Nat Biotechnol._ 2022 Apr 10. doi: [10.1038/s41587-023-01733-8](https://doi.org/10.1038/s41587-023-01733-8).
```
grst, line 48:
```suggestion
```

grst:
(instance repo test is failing because this PR is from a fork -- we can just merge it regardless)

flying-sheep:
Yeah, but how do we fix that? People who aren’t maintainers/owners of this repo should probably not see this test fail by default. We need to think about the security implications of this too.

flying-sheep (changes requested):

flying-sheep (approved):

PR #168, opened 2023-04-27
Cruft PR workflow

### LINKED ISSUE #138: Template sync fails when changes are made to workflow files
In a test-run, the template sync failed since apparently
the `sync.yaml` action is not permitted to make changes to files in the
`.github/workflows` directory:
https://github.com/icbi-lab/infercnvpy/actions/runs/3875431171/jobs/6607954535

This is a bit of a problem for which we still need to figure out a
solution. One of them is to use a personal access token. Another one
could be to centralize template sync, which I would have liked to avoid.

In #137, I roll back the changes made to workflow files since the last release,
as a short-term fix.

flying-sheep:
Derived from https://gist.github.com/grst/7593e7ff965df9a6aeab5c4fbdd796bf

Close #138


.github/workflows/cruft-prs.yml
```diff
@@ -0,0 +1,39 @@
+name: make cruft PRs for all projects using us
+on:
+  release:
```
grst, line 3: I'd also add workflow_dispatch

flying-sheep: Hmm, then we’d have to add all the release metadata used in the PR title body as `input`s

flying-sheep: OK, with the script it’s easy

flying-sheep:
blocked by https://github.com/charliermarsh/ruff/issues/4134

scripts/src/scverse_template_scripts/cruft_prs.py
```diff
@@ -0,0 +1,171 @@
...
+    return True
+
+
+def make_pr(con: GitHubConnection, release: GHRelease, repo_url: str) -> None:
+    pr = PR(release)
+    log.info(f"Sending PR to {repo_url}: {pr.title}")
+
+    # create fork, populate branch, do PR from it
+    origin = con.gh.get_repo(repo_url.removeprefix("https://github.com/"))
+    repo = origin.create_fork()
```
grst, line 142: does this work if the fork already exists? 

Do we need to pull upstream into the fork if it already exists from the last update?

scripts/src/scverse_template_scripts/cruft_prs.py
```diff
@@ -0,0 +1,171 @@
...
+    with TemporaryDirectory() as td:
+        updated = cruft_update(con, repo, Path(td), pr)
+    if updated:
+        origin.create_pull(pr.title, pr.body, ...)  # TODO
```
grst, line 146: what happens if a pull request already exists from the last update but hasn't been merged yet?

scripts/src/scverse_template_scripts/cruft_prs.py
```diff
@@ -0,0 +1,171 @@
...
+        return f"Update template to {self.release.tag_name}"
+
+    @property
+    def branch(self) -> str:
+        return f"template-update-{self.release.tag_name}"
+
+    @property
+    def body(self) -> str:
+        template_usage = "https://cookiecutter-scverse-instance.readthedocs.io/en/latest/template_usage.html"
+        return f"""\
```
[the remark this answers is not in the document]
flying-sheep: since we insert the body, we need to `indent` that one. and once we use that, a refactor could break things as the `indent`, would need a different argument. I’d rather use a template file than that.

scripts/src/scverse_template_scripts/cruft_prs.py
flying-sheep: great point, we should add a test for that

scripts/src/scverse_template_scripts/cruft_prs.py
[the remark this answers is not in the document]
grst: ah got it!

scripts/src/scverse_template_scripts/cruft_prs.py
flying-sheep: hm, testing the GitHub API doesn’t seem to be possible without mocking everything manually

scripts/src/scverse_template_scripts/cruft_prs.py
flying-sheep: done

scripts/pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+[tool.hatch.envs.test.scripts]
+test = "pytest"
+
+[tool.pytest.ini_options]
+addopts = [
+    "--import-mode=importlib",
+    "-pgit",
+    "-pnetwork", "--disable-network",
+    "-presponsemock"
+]
```
grst, line 57: why is this needed?

scripts/pyproject.toml
```diff
@@ -0,0 +1,108 @@
...
+    "S",
+    "T",
+    "TID",
+    "UP",
+    "W",
+    "YTT",
+]
+ignore = [
+    "S101",  # assert should be allowed
+    "S603",  # suprocess with shell=False should be allowed
```
grst, line 97:
```suggestion
    "S603",  # subprocess with shell=False should be allowed
```

grst (approved):
Looks beautiful! 
Now how do we best go about merging this and testing this in a real world setting? 

In general, I would suggest that the next release of the cookiecutter template *only* includes the changes from this PR (i.e. delete `sync.yaml`). This makes it easier to communicate to the template users that we switched to a new system and the respective PRs will be easy to review and merge (there shouldn't be any merge conflicts).

scripts/pyproject.toml
flying-sheep: The `-p???` switches activate the pytest plugins used and make pytest complain if they’re missing. The `--disable-network` switch makes sure all network requests are mocked out.

grst:
merge this PR and cherry-pick it in the `v0.2.x` branch, then make a release and see what happens?

Or is there a better way  of testing your "on release" action before actually making a release?

flying-sheep:
You’re right!

PR #176, opened 2023-04-28
Better instance repo sync

### LINKED ISSUE #91: Improvements to template sync
- [ ] Close a PR in the instance repo instead of merging it. Instead, push to the main branch of the instance repo on every push to the main branch of the template repo. This ensures the instance never gets out of sync, e.g. because a PR fails to be merged due to a merge conflict. 
- [ ] Currently instance builds don't work for PRs from forks because they don't get access to the scverse-bot token stored as secret. The "best" practice seems to be to run an action on `pull_request` that builds the template, and stores the output as artifact. Then in a separate action on `workflow_run` (which has access to the secrets), obtain the artifact and do everything needed to push to the instance repo. This approach is described [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

grst:
- Update actions to close PRs on merge, but push to main of instance repo instead
- Fix HEREDOC in pos-gen hook

Partly addresses #91

### DIFF
--- a/.github/workflows/deploy-instance-pr-actions.yml
+++ b/.github/workflows/deploy-instance-pr-actions.yml
@@ -26,7 +26,7 @@ jobs:
           )
           echo "SISTER_PR_ID=$PR_ID" >> $GITHUB_ENV
 
-      - name: Close or merge sister PR
+      - name: When closing or merging also close sister PR
         if: ${{ env.SISTER_PR_ID != 'NA' }}
         run: |
           gh pr close $SISTER_PR_ID \
--- a/.github/workflows/deploy-instance-pr-actions.yml
+++ b/.github/workflows/deploy-instance-pr-actions.yml
@@ -26,17 +26,10 @@ jobs:
           )
           echo "SISTER_PR_ID=$PR_ID" >> $GITHUB_ENV
 
-      - name: Close sister PR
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-instance-repo.yml
```diff
@@ -85,7 +93,10 @@ jobs:
           force: false
           directory: ${{ env.INSTANCE_REPO }}
 
+      # Run the folllwing steps only on the pull request event -- in the other case we just commited
```
flying-sheep, line 96:
```suggestion
      # Run the following steps only on the pull request event -- in the other case we just commited
```

.github/workflows/deploy-instance-pr-actions.yml
```diff
@@ -26,19 +26,12 @@ jobs:
           )
           echo "SISTER_PR_ID=$PR_ID" >> $GITHUB_ENV
 
-      - name: Close sister PR
-        if: ${{ env.SISTER_PR_ID != 'NA' && github.event.action == 'closed' && !github.event.pull_request.merged }}
+      - name: Close or merge sister PR
```
flying-sheep, line 29:
```suggestion
      - name: When closing or merging also close sister PR
```

flying-sheep (approved):
some clarifications would be nice

PR #182, opened 2023-04-28
Run tests on a bi-weekly basis

grst:
the idea here is to catch errors with pre-releases, even if no PR is opened. 

Currently this is set to a weekly schedule (5am on Sundays).


grst:
@flying-sheep / @Zethson do we want to include this?

Zethson:
I get the idea and maybe weekly is a bit too often? Biweekly or monthly?

grst:
how long are pre-releases usually out before the actual release?

Zethson:
There's no rule here I guess.

grst:
for pandas, they announced it with "two weeks", but it ended up being 5

Zethson:
I mean if a package isn't actively being developed, the chance of people looking at CI failures is rather minuscule haha.

I'm fine with weekly, but if you already have failing CI (which surprisingly many packages do quite often cough scanpy) you might be hella annoyed.

flying-sheep:
> for pandas, they announced it with "two weeks", but it ended up being 5

Wow, two weeks is nothing. I think 5 is realistic for us noticing and having time to work on a fix.

flying-sheep:
I see, so this will run it twice around the change of month and once in the middle: <https://crontab.guru/#0_5_*/14_*_*>

> next at 2023-06-15 05:00:00
> **then at 2023-06-29 05:00:00
> then at 2023-07-01 05:00:00**
> then at 2023-07-15 05:00:00
> then at 2023-07-29 05:00:00

can we do something like “on first and on 14th” instead?

.github/workflows/test.yaml
```diff
@@ -5,6 +5,8 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  schedule:
+    - cron: "0 5 */14 * *"
```
flying-sheep, line 9:
```suggestion
    - cron: "0 5 1,15 * *"
```

This would be a better bi-weekly: <https://crontab.guru/#0_5_1,15_*_*>

> next at 2023-06-15 05:00:00
> then at 2023-07-01 05:00:00
> then at 2023-07-15 05:00:00
> then at 2023-08-01 05:00:00
> then at 2023-08-15 05:00:00

flying-sheep (approved):

PR #205, opened 2023-06-05
Update accepted docutils versions

### LINKED ISSUE #190: "references" page in documentation not rendered correctly
The footer appears on the side: 

image

I have the same problem in scirpy after adapting the template, so this is not an isolated issue of the instance repo.

flying-sheep:
Fixes  #190

Docutils 0.20 has the patch we need, and Sphinx 7.0.1 is necessary for docutils 0.20, sphinx-book-theme only accepts sphinx <7

Therefore we can specify `docutils>=0.8, !=0.18.*, !=0.19.*`, which will give us docutils 0.17 until sphinx-book-theme supports sphinx 7.

### DIFF
--- a/{{cookiecutter.project_name}}/docs/conf.py
+++ b/{{cookiecutter.project_name}}/docs/conf.py
@@ -57,6 +57,7 @@ extensions = [
     "sphinx_autodoc_typehints",
     "sphinx.ext.mathjax",
     "IPython.sphinxext.ipython_console_highlighting",
+    "sphinxext.opengraph",
     *[p.stem for p in (HERE / "extensions").glob("*.py")],
 ]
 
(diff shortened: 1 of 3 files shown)


grst:
>  which will give us docutils 0.17 until sphinx-book-theme supports sphinx 7

what's the benefit of 0.17 over .18 and .19?

flying-sheep:
It creates valid HTML and doesn’t mess up our layout horribly. See #190

flying-sheep:
looks like it yeah!

flying-sheep:
@grst do you want to do the release with this? I hope the update code finally works!

flying-sheep:
yeah, I’ve been doing a lot of releases because there was no good way to test

PR #211, opened 2023-06-26
Add TOML  formatting

flying-sheep:
The hack being present makes pre-commit hang infinitely on my system, so we need a better solution.


grst (approved):
LGTM

grst:
Some tests still not passing though?

flying-sheep:
Ah, I just checked the maintenance status and it’s actually not maintained. So we need a different solution…

PR #224, opened 2023-08-11
Make test part of the  build and release workflow. Run tests on all branches.

bernheder:
I think the `test` job should be part of both the `build` and `release` workflows, as the package should only be built and released if it functions correctly.

Furthermore, since the test job is  then part of the build/release process and consistently runs on the same trigger as the `build` workflow (push/pull requests to the main branch), it can either be removed or, in this case, modified to run in different scenarios (I would suggest running tests on all branches).

### DIFF
--- a/{{cookiecutter.project_name}}/.github/workflows/build.yaml
+++ b/{{cookiecutter.project_name}}/.github/workflows/build.yaml
@@ -11,8 +11,54 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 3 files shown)


grst:
I don't think we need this...
 * tests already run on all branches for which a pull request exists (and a pull request is the only way to get something in the main branch) 
 * anything in the main branch needed to pass the tests to get in, so we don't need to rerun them on release.

PR #225, opened 2023-08-11
TOML formatting using taplo

### LINKED ISSUE #212: TOML formatting
This question still doesn’t seem exhaustively answered: https://github.com/toml-lang/toml/issues/532

The most promising option right now seems to be: https://github.com/tamasfe/taplo

flying-sheep:
Fixes #212, finally!

an alternative to #299, see #212 for a comparison (taplo is now directly published to PyPI, which makes this hook a nice, perfectly integrated alternative)

Outstanding questions:
1. should we customize taplo more than just with the spaces? E.g. “never collapse multiline arrays” might be desirable. For reference here: https://github.com/tamasfe/taplo/issues/358
2. do we prefer #299?


flying-sheep:
Closing in favor of https://github.com/scverse/cookiecutter-scverse/pull/299, see comments there

PR #244, opened 2023-11-28
Use `coverage run`

### LINKED ISSUE #239: Run tests using `coverage run`
Currently neither `pytest --cov` nor `coverage run -m pytest` come without drawbacks:

- `pytest --cov` only starts collecting coverage after plugins have been loaded, so it doesn’t correctly collect coverage in some situations. `coverage run` is just more reliable.
- But `covtest run` doesn’t have pytest’s behavior of not adding the working directory to `sys.path` and can’t be made to do that: https://github.com/nedbat/coveragepy/issues/1696

I think due to our use of the `src` layout, the latter is the better tradeoff

flying-sheep:
Fixes  #239

Adds

image


grst (approved):
LGTM

PR #247, opened 2023-11-28
Conditionally render dev docs

### LINKED ISSUE #173: Separate template documentation from packages' developer documentation
Right now we have `template_usage.md` and `contributing.md`. 
The former documents details about the template, the latter is targeted at developers of an actual package using the template. 

Currently, both documents are included in an instance of the template and rendered on readthedocs. 

I think we should exclude `template_usage` from instances. However, we then need a place where documentation for the template is hosted. Ideas: 

 * Just stick to github markdown files
 * find a way to **only** include it in the `cookiecutter-scverse-instance` repo, but not other instances
 * have another documentation website dedicated to the template

flying-sheep:
Fixes #173

### [Rendered](https://cookiecutter-scverse-instance--100.org.readthedocs.build/en/100/)


hooks/post_gen_project.py
```diff
@@ -1,15 +1,15 @@
-#!/bin/sh
+#!/bin/env python
```
grst, line 1: This doesn't run in an environment, so on some systems the default python might still point to python2

```suggestion
#!/bin/env python3
```

hooks/pre_gen_project.py
```diff
@@ -0,0 +1,4 @@
+#!/bin/env python
```
grst:
```suggestion
#!/bin/env python3
```

grst (approved):
LGTM!

PR #249, opened 2023-11-28
Change workflow and documentation to trusted publishing

### LINKED ISSUE #228: Switch to trusted publishing
It’s easier than wrangling tokens.

1. Switch the publish workflow to this: https://github.com/flying-sheep/hatch-docstring-description/blob/39ef405c98ed8b9584db9f591398fcf2c67437d7/.github/workflows/pub.yml
2. Add to the docs that users need to
   1. add an environment to their repo called `pypi` (no settings, just have it)
   2. enter `username`, `repo`, `pub.yml` and `pypi` in a form and submit it (described [here](https://docs.pypi.org/trusted-publishers/))

grst:
Close #228

### DIFF
--- a/{{cookiecutter.project_name}}/.github/workflows/release.yaml
+++ b/{{cookiecutter.project_name}}/.github/workflows/release.yaml
@@ -1,9 +1,8 @@
 name: Release
 
 on:
-  push:
-    tags:
-      - "*.*.*"
+  release:
+    types: [published]
 
 # Use "trusted publishing", see https://docs.pypi.org/trusted-publishers/
 jobs:
--- a/{{cookiecutter.project_name}}/.github/workflows/release.yaml
+++ b/{{cookiecutter.project_name}}/.github/workflows/release.yaml
@@ -18,7 +18,7 @@ jobs:
      - uses: actions/checkout@v3
      - uses: actions/setup-python@v4
        with:
-          python-version: "3.11"
+          python-version: "3.x"
           cache: "pip"
       - run: pip install build
       - run: python -m build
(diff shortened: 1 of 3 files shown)


{{cookiecutter.project_name}}/.github/workflows/release.yaml
```diff
@@ -5,26 +5,23 @@ on:
...
-        run: pip install hatch
-
-      - name: Build project for distribution
-        run: hatch build
-
-      - name: Publish a Python distribution to PyPI
+     - uses: actions/checkout@v3
+     - uses: actions/setup-python@v4
+       with:
+          python-version: "3.11"
```
flying-sheep, line 22:
```suggestion
          python-version: "3.x"
```

only needs build deps, which shouldn’t be as sensitive as runtime deps like numba

{{cookiecutter.project_name}}/.github/workflows/release.yaml
```diff
@@ -5,26 +5,23 @@ on:
...
-          python-version: "3.10"
-
-      - name: Install hatch
-        run: pip install hatch
-
-      - name: Build project for distribution
-        run: hatch build
-
-      - name: Publish a Python distribution to PyPI
+     - uses: actions/checkout@v3
```
flying-sheep, line 19:
```suggestion
     - uses: actions/checkout@v4
       with:
         filter: blob:none
         fetch-depth: 0
```

flying-sheep (changes requested):

flying-sheep (approved):

PR #263, opened 2024-01-24
Fix copy without render

grst:
`release.yaml` actually contains a jinja2 variable, so it should not be listed in "copy without render"


flying-sheep (approved):
You spooked me. Luckily it’s just a kinda cosmetic use, not breaking anything

PR #278, opened 2024-04-23
Use cookiecutter variable for project repo in conf.py

Zethson:


flying-sheep (approved):

flying-sheep:
Oof actually I think this isn’t correct.

`html_context["github_repo"]` should only contain the repo name, not the URL

PR #284, opened 2024-05-18
Update required python versions according to scientific python SPEC 0

grst:


Zethson (approved):
Gotta say, SPEC 0 is pretty aggressive with minimum versions, but it's probably a good thing

PR #299, opened 2024-07-30
TOML formatting using pyproject-fmt

### LINKED ISSUE #212: TOML formatting

flying-sheep:
Fixes  #212

An alternative to #225, see #212 for a comparison

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -1,7 +1,0 @@
-{
-    "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
-    "formatter": {
-        "indentStyle": "space",
-        "indentWidth": 4
-    }
-}
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -1,0 +1,7 @@
+{
+    "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+    "formatter": {
+        "indentStyle": "space",
+        "indentWidth": 4
+    }
+}
--- a/{{cookiecutter.project_name}}/pyproject.toml
+++ b/{{cookiecutter.project_name}}/pyproject.toml
@@ -10,39 +10,39 @@ readme = "README.md"
 requires-python = ">=3.10"
 license = {file = "LICENSE"}
 authors = [
-    {name = "{{ cookiecutter.author_full_name }}"},
+  {name = "{{ cookiecutter.author_full_name }}"},
 ]
 maintainers = [
-    {name = "{{ cookiecutter.author_full_name }}", email = "{{ cookiecutter.author_email }}"},
+  {name = "{{ cookiecutter.author_full_name }}", email = "{{ cookiecutter.author_email }}"},
 ]
 urls.Documentation = "https://{{ cookiecutter.project_name }}.readthedocs.io/"
 urls.Source = "{{ cookiecutter.project_repo }}"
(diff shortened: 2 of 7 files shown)


biome.jsonc
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "https://biomejs.dev/schemas/1.8.3/schema.json",
+    "formatter": {
+        "indentStyle": "space",
+        "indentWidth": 4
+    }
+}
```
flying-sheep, line 7: we could also leave it out, then all JSON files get formatted with tabs

grst: Let's keep it, too bad biome doesn't support editorconfig.

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -1,139 +1,149 @@
...
-    # First line should be in imperative mood; try rephrasing
-    "D401",
-    ## Disable one in each pair of mutually incompatible rules
-    # We don’t want a blank line before a class docstring
-    "D203",
-    # We want docstrings to start immediately after the opening triple quote
-    "D213",
+[tool.coverage.run]
+source = [ "{{ cookiecutter.package_name }}" ]
+omit = [
```
grst, line 135: Are all these rearrangements necessary? This will create one huge merge conflict when making the template-update PRs

grst:
I like it in principle! Shall we wait a bit longer until the biome-pre-commit and pyproject-fmt are maturing a bit more?
How likely is it that pyproject-fmt will become part of ruff?

{{cookiecutter.project_name}}/pyproject.toml
flying-sheep: That’s the thing: pyproject-fmt dictates the order of these blocks.

If it was possible to run the hooks before creating the diff, maybe we could avoid conflicts

flying-sheep:
> Shall we wait a bit longer until the biome-pre-commit and pyproject-fmt are maturing a bit more?

I think pyproject-fmt is mature enough, for biome-pre-commit we should maybe wait until someone replies to my comment in https://github.com/biomejs/pre-commit/issues/5

> How likely is it that pyproject-fmt will become part of ruff?

it’s a completely separate rust project, and targets a different type of file, so maybe not that likely?

but then again the ruff people have adopted `uv` and `rye` so who knows!

biome.jsonc
flying-sheep: It’s coming, actually! https://github.com/biomejs/biome/issues/1724

flying-sheep:
I’ll split this up into two PRs

grst (approved):
I think I'd prefer this one over taplo - In addition to the reasons you mentioned it
 * doesn't require another `.<formatter>` config file
 * seems to do more stuff that's specific to pyproject.toml, but useful (such as sorting dependencies)

flying-sheep:
OK! Do you think we should leave it unconfigured (2 spaces indent) or add

```toml
[tool.pyproject-fmt]
indent = 4
```

flying-sheep:
Welp, we should decide on indents before next release, but I’m going to merge this

PR #307, opened 2024-09-02
Fix repo metadata

flying-sheep:
It’s used here: https://github.com/pydata/pydata-sphinx-theme/blob/15494ecd48147e03d57fd462c1eaf8fb5aeb67d2/src/pydata_sphinx_theme/edit_this_page.py#L51-L55

Alternatively, we could define `__repo_url` for a more DRY cookiecutter (double underscore means that the user doesn’t get asked, but {{substitution}} still happens)


Zethson (approved):
Thank you for fixing this. I'm fine with this. Verbose but explicit.

PR #308, opened 2024-09-02
Hatch envs

flying-sheep:
TODO:

- [ ] Maybe mention `hatch env remove` or `hatch env prune`?

### [Rendered docs](https://cookiecutter-scverse-instance--145.org.readthedocs.build/en/145/contributing.html)

### DIFF
--- a/{{cookiecutter.project_name}}/pyproject.toml
+++ b/{{cookiecutter.project_name}}/pyproject.toml
@@ -60,6 +60,18 @@ packages = [ "src/{{ cookiecutter.package_name }}" ]
 
 {% endif -%}
 
+[tool.hatch.envs.default]
+installer = "uv"
+features = [ "dev" ]
+
+[tool.hatch.envs.docs]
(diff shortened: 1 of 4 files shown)


{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -60,6 +61,19 @@ packages = [ "src/{{ cookiecutter.package_name }}" ]
 
 {% endif -%}
 
+[tool.hatch.envs.default]
+installer = "uv"
```
Zethson, line 65: Naive question: What happens if `uv` is not available?

Zethson (approved):

{{cookiecutter.project_name}}/pyproject.toml
flying-sheep: [Mu](https://csarven.ca/mu), Hatch comes with an internal version of uv, so it’s always available.

PR #311, opened 2024-09-05
Update post-gen output for hatch

flying-sheep:


hooks/post_gen_project.py
```diff
@@ -42,11 +42,13 @@
 ┃                         \x1b[1mInstall the package\x1b[0m                          ┃
```
flying-sheep, line 42: hm, maybe this title is no longer correct.

PR #312, opened 2024-09-05
Add VS Code integration

flying-sheep:
This makes it easier to teach this as everything is already set up when people use VS Code. Maybe the variable should be a multiple choice for “none”, “vs code”, “all”, with the ability to extend support for PyCharm and so on?


{{cookiecutter.project_name}}/.gitignore
```diff
@@ -3,6 +3,13 @@
 *~
 buck-out/
 
+{% if not cookiecutter.ide_integration -%}
```
Zethson, line 6: Out of convenience, I would just always add these to the `.gitignore`. Even if the person setting up the project might not be using them, other contributors will.

{{cookiecutter.project_name}}/.gitignore
```diff
@@ -3,6 +3,13 @@
...
+# IDEs
+/.idea/
```
Zethson, line 8: This is a very technical detail that is not important but there are a few files in the `.idea` folder that Jetbrains thinks are fine to commit. But let's ignore this.

{{cookiecutter.project_name}}/.gitignore
Zethson: Oh wait I think I got the logic the wrong way around. Now we're keeping them, right?

Zethson (approved):

{{cookiecutter.project_name}}/.gitignore
flying-sheep: I always commit them actually.

{{cookiecutter.project_name}}/.gitignore
flying-sheep: yeah, that’s what I mean with

> Maybe the variable should be a multiple choice for “none”, “vs code”, “all”, with the ability to extend support for PyCharm and so on?

PR #328, opened 2024-10-21
fix hatch snippets for docs build in contributing.md

eroell:
I think that was even found by @jan-engelmann :D


jan-engelmann:
thanks @eroell 

I think the docs would greatly benefit of a little introduction section on hatch and what needs to be done such that hatch python and not conda is used. I know that this is hatch and not cookiecutter specific but I think it would help adoption.

eroell:
agree with Jan. for me (hatch noob) the motivation and an end-to-end setup are not there clear enough either

flying-sheep (approved):

flying-sheep:
Could you please open an issue for that? This PR just fixes some typos and is therefore not an ideal place to discuss bigger changes!

PR #356, opened 2025-03-05
fix test workflow badge link

le-ander:
The current link of the test badge leads to a "workflow not found" page. This is because of a mismatch in the extension (`.yml` instead of `.yaml`)


le-ander:
Also mambaforge seems to be deprecated according to: https://github.com/conda-forge/miniforge
Replaced the pointer to mambaforge with miniforge.

Personally, I'm not a big fan of using conda and would prefer if a more vanilla non-conda way was recommended here.

Issue #54, opened 2022-05-10
Better doc strings


ivirshup:
We should be demonstrating more features in the docs.

In particular, I think we should be demonstrating what a fully featured docstring looks like. This means adding type hints, parameter docs, and ideally examples.

PR #365, opened 2025-03-18
Add more elaborate example

Zethson:
Fixes https://github.com/scverse/cookiecutter-scverse/issues/54

Adds a more elaborate example to the preprocessing code that highlights typehints, examples, keyword only arguments, and more.

We could add an even crazier example that doesn't use Any, improves the handling if missing MuData and SpatialData installations, ... but I think then we overwhelm people. I am hoping to hit the right sweet spot.

### DIFF
--- a/{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
+++ b/{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
@@ -64,7 +64,7 @@ def elaborate_example(
     >>> elaborate_example(
     ...     [adata, mudata, spatial_data],
     ...     lambda vals: f"Statistics: mean={vals.mean():.2f}, max={vals.max():.2f}",
-    ...     {"var_key": "CD45", "modality": "rna", "min_value": 0.1}
+    ...     {"var_key": "CD45", "modality": "rna", "min_value": 0.1},
     ... )
     ['Statistics: mean=1.24, max=8.75', 'Statistics: mean=0.86, max=5.42']
     """
--- a/{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
+++ b/{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
@@ -25,6 +25,7 @@ def basic_preproc(adata: AnnData) -> int:
     print("Implement a preprocessing function here.")
     return 0
 
+
 def elaborate_example(
     items: Iterable[ScverseDataStructures],
     transform: Callable[[Any], str],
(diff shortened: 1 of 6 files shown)


{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -15,3 +24,68 @@ def basic_preproc(adata: AnnData) -> int:
     """
     print("Implement a preprocessing function here.")
     return 0
+
+
+def elaborate_example(
+    items: Iterable[ScverseDataStructures],
+    transform: Callable[[Any], str],
+    /,  # indicates the end of positional only arguments
```
grst, line 32: There are very few cases where positional-only arguments make sense. Not sure I'd show this here. 
```suggestion

```

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -1,5 +1,14 @@
+from collections.abc import Callable, Iterable
+from typing import Any, TypeVar
+
+import numpy as np
 from anndata import AnnData
 
+MuData = TypeVar("MuData")
```
grst, line 7: Why is TypeVar required here? Can't you use MuData/AnnData directly? 

or

```
if TYPE_CHECKING:
   from mudata import MuData
   from spatialdata import SpatialData
```

if you are worried about additional imports?

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -15,3 +24,68 @@ def basic_preproc(adata: AnnData) -> int:
...
+def elaborate_example(
+    items: Iterable[ScverseDataStructures],
+    transform: Callable[[Any], str],
+    /,  # indicates the end of positional only arguments
+    *,  # functions after the asterix are key word only arguments
+    layer_key: str | None = None,
+    mudata_mod: str | None = "rna",  # Only specify defaults in the signature, not the docstring!
+    sdata_table_key: str | None = "table1",
+    max_items: int = 100,
+) -> list[Any]:
```
grst, line 38: given the type of `transform` this would return a list of `str` wouldn't it?

grst (commented):
I like it, added some comments

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
Zethson: Ah right, I had it earlier and then forgot that I had specified the transform type more narrowly. Thanks

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
Zethson: Unless I'm stupid, we'd then also require some patching for sphinx (mock imports or nitpick ignores etc). I'd like to avoid that for MuData and SpatialData because people might not notice that we did so. I think it's easier if we keep it as I had it.

grst: ok, fair enough

grst (approved):

PR #366, opened 2025-03-18
Switch to session-info2

Zethson:
Fixes https://github.com/scverse/cookiecutter-scverse/issues/363


{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -23,67 +23,22 @@ body:
...
-        Please paste below the output of
+        Which version of packages.
+
+        Please install `session-info2`, run the following command in a notebook,
+        click the “Copy as Markdown” button, then paste the results into the text box below.
 
         ```python
-        import session_info
-        session_info.show(html=False, dependencies=True)
+        In[1]: import session_info2; session_info2.session_info(dependencies=True)
```
Zethson, line 34: @flying-sheep do we need to import the package here before we run the rest? Yes, right? While possible, it'd be a bit annoying to get the cookiecutter variable in here.

flying-sheep: With `dependencies=True` not necessarily, but you should paste what it does into the `placeholder` anyway instead of removing it, so people can see how it’s supposed to look like

Zethson: This is adapted from anndata:
https://raw.githubusercontent.com/scverse/anndata/refs/heads/main/.github/ISSUE_TEMPLATE/bug-report.yml

Still want to keep the placeholder paste?

flying-sheep: I think it probably helps, what do you think?

{{cookiecutter.project_name}}/.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -23,67 +23,74 @@ body:
...
+        setuptools	76.1.0
+        Deprecated	1.2.18
+        packaging	24.2
+        python-dateutil	2.9.0.post0
+        ----	----
+        Python	3.13.2 | packaged by conda-forge | (main, Feb 17 2025, 14:10:22) [GCC 13.3.0]
+        OS	Linux-6.11.0-109019-tuxedo-x86_64-with-glibc2.39
+        Updated	2025-03-18 15:47
+    validations:
+      required: true
```
grst, line 96: You really want to force this? I'd still rather have users to submit incomplete bug reports than not reporting them at all.

grst (approved):

PR #368, opened 2025-03-18
Add hatch matrix

### LINKED ISSUE #338: pyproject.toml missing hatch matrix definition
Hi,
I think that `pyproject.toml` could benefit from the addition of a hatch matrix definition, such as:
```toml
[[tool.hatch.envs.hatch-test.matrix]]
python = ["3.10", "3.11", "3.12"]
features = ["test"]
```

At least for me, running tests locally with `hatch test` fails without this addition, because hatch considers my active Python version to be compatible and doesn't launch into a testing environment but just uses my local environment. In fact, I am not sure [these two lines](https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/pyproject.toml#L75-L76) have any effect while testing. In the [docs](https://hatch.pypa.io/1.13/tutorials/testing/overview/) they use the matrix definition, although for granular dependency management they do seem to use `tool.hatch.envs.hatch-test` (i.e. not matrix) ([see here](https://hatch.pypa.io/1.13/config/internal/testing/#default-arguments)).

Zethson:
Fixes #338 

I am not sure whether we want to do this because users would have to take care of this as well and update it in addition to Github Actions workflows etc.

The default behavior is somewhat sane in my opinion where it:

> If no environment options are selected, the test command will only run tests in the first defined environment that either already exists or is compatible. Additionally, the checking order will prioritize environments that define a [version of Python](https://hatch.pypa.io/1.13/config/environment/overview/#python-version) that matches the interpreter that Hatch is running on.

https://hatch.pypa.io/1.13/tutorials/testing/overview/

@flying-sheep what is your opinion?

### DIFF
--- a/{{cookiecutter.project_name}}/pyproject.toml
+++ b/{{cookiecutter.project_name}}/pyproject.toml
@@ -72,8 +72,9 @@ scripts.build = "sphinx-build -M html docs docs/_build {args}"
 scripts.open = "python -m webbrowser -t docs/_build/html/index.html"
 scripts.clean = "git clean -fdX -- {args:docs}"
 
-[tool.hatch.envs.hatch-test]
-features = [ "test" ]
+[[tool.hatch.envs.hatch-test.matrix]]
+python = ["3.10", "3.11", "3.12", "3.13"]
+features = ["test"]
 
 [tool.ruff]
 line-length = 120


flying-sheep:
This does nothing, see #338

grst:
So what's the recommended way then to use hatch to test against multiple python versions?

flying-sheep:
https://github.com/scverse/cookiecutter-scverse/issues/338#issuecomment-2740078839

> \[we are not missing a matrix definition since] the [default one](https://hatch.pypa.io/1.13/config/internal/testing/#matrix) exists and ~~I think it~~ works fine

rely on the default matrix, as explained in our docs: https://cookiecutter-scverse-instance.readthedocs.io/en/latest/contributing.html#writing-tests

> Alternatively, you can run all tests from the command line by executing
> 
> ```bash
> hatch test --all  # test with all supported Python versions
> ```

I guess we could add a link explaining `hatch test -py 3.11`.

PR #370, opened 2025-03-19
Suggest to use `uv` to run cruft

grst:


README.md
```diff
@@ -62,13 +60,26 @@ updates to your project whenever a new template version is released.
...
 This will create a git repository generated from the template.
 Now `cd` into the newly created directory and make the initial commit!
 Don't forget to create a repository on GitHub and upload your project.
 
+<details>
+<summary>Working without `uv`</summary>
+If you prefer to manage the dependencies yourself, you can run
+
+```bash
+pip install cruft pre-commit
```
flying-sheep, line 76: maybe rephrase as “if you want to permanently install the tools”

and tell them to use `uv tool install` or `pipx install`?

grst: I was more thinking about someone who doesn't want to or cannot use any additional tools... but since `uv` itself can be installed with pip it's not a good argument. 

So I'm fine with that

flying-sheep (approved):

PR #375, opened 2025-03-27
Use uv caching

flying-sheep:
We use Hatch as an env manager and uv as an installer (partially through hatch, partially by itself)

The `pypa/hatch@install` action doesn‘t set up caching, yet everything we do benefits from it:
- the test job can really benefit from this, if a user has a dependency that needs to be built, having it in the `uv` cache really speeds up tests
- `pyproject-build --installer=uv` as well as `uv build` (which is shorter, so maybe the way to go) both use `uv` to install the build deps. The benefit isn’t large (there aren’t that many build deps) but why not be consistent?

This PR changes #373 so it uses uv caching everywhere, and installs `hatch` via `uv` while we’re at it.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -73,8 +73,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Install uv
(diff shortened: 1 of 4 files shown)


grst:
LGTM

PR #377, opened 2025-03-31
Even more precautionary bash settings

grst:
To follow up on https://github.com/scverse/cookiecutter-scverse/pull/373#discussion_r2016041818

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,7 +8,7 @@ on:
 
 defaults:
   run:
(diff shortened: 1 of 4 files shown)


flying-sheep (approved):

flying-sheep:
Sure. In the end, bash is simply not possible to isolate against silent failure though.

E.g. `somevar=$(some-command)` can still fail silently with `-euo pipefail`, one would have to use `some-command | read -r somevar`.

PR #380, opened 2025-04-03
Fix repo migration commands

le-ander:
I just migrated an existing repo to the template, following the "Migrate existing projects to using this template" steps. I found that cruft creates a folder with the project name inside `./template/` rather than writing directly into `./template/`. Also, after step 4. the user will be inside `$REPO_cookiecutterized`.

I have updated the rsync command in 5. to account for these two things. Hope that makes sense.


grst (approved):
Thanks!

PR #383, opened 2025-04-05
Delete docs/Makefile

grst:
How to build docs is defined via hatch in `pyproject.toml`

This Makefile is the default from sphinx, but I don't think we should keep it anymore as the pyproject.toml is the single point of truth


flying-sheep (approved):

.github/workflows/test.yaml
```diff
@@ -64,15 +64,13 @@ jobs:
...
           python -c "import ${{ matrix.package-name }}"
       # Docs
       - name: Build the documentation
-        env:
-          SPHINXOPTS: -W --keep-going
         run: |
-          cd "$PROJECT_ROOT/docs"
-          make html
+          cd $PROJECT_ROOT
+          hatch run docs:build -W
```
flying-sheep, line 73: this will build with uv, so it won’t use any of the caches we set up (`pip`, `pre-commit`)

so to speed up our CI, we should switch this workflow to uv as well.

PR #384, opened 2025-04-07
Switch root tests to uv

flying-sheep:
We should probably merge this before #383


.github/workflows/test.yaml
```diff
@@ -32,24 +32,17 @@ jobs:
...
       - name: Build from template
-        shell: python
         run: |
-          from cookiecutter.main import cookiecutter
-          overrides = dict(package_name='${{ matrix.package-name }}')
-          cookiecutter('.', no_input=True, extra_context=overrides)
+          uvx --with pre-commit cruft create  . --no-input --extra-context '{"package_name": "${{ matrix.package-name }}"}'
       # Check
       - name: Set up pre-commit cache
         uses: actions/cache@v4
```
grst, line 48: Does this still do anything? I'd guess yes, because I don't think pre-commit uses `uv` to create environments?

flying-sheep: unless we also install `pre-commit-uv` which we could: https://github.com/tox-dev/pre-commit-uv

grst: Let's do that. It's maintained by tox and if it breaks the cost is low because it's just in the main repo, not the template.

.github/workflows/test.yaml
```diff
@@ -37,12 +37,15 @@ jobs:
...
       - name: Install Ubuntu system dependencies
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get install pandoc
-      - name: Install build & check dependencies
-        run: python -m pip install --upgrade pip wheel cookiecutter pre-commit
+      - name: Install build dependencies
+        run: uv pip install --system cookiecutter
       # Build
       - name: Build from template
         shell: python
```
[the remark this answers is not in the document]
flying-sheep: Neat, only using CLIs sure is better.

I think we misunderstood how `pre-commit-uv` works though (it patches `pre-commit`), so I switched back to installing the tools instead of using them inline for clarity.

grst (approved):

PR #390, opened 2025-04-09
Get test matrix from hatch

grst:


flying-sheep:
Oooh that’s sexy, I didn’t think of that!

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -18,44 +18,61 @@ defaults:
...
+        os: [ubuntu-latest]
+        include: ${{ fromJson(needs.get-environments.outputs.matrix) }}
+
+    name: ${{ matrix.name }}
 
     steps:
       - uses: actions/checkout@v4
-        with:
-          filter: blob:none
-          fetch-depth: 0
```
flying-sheep, line 47: why this?

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -18,44 +18,61 @@ defaults:
...
-            pip-flags: "--pre"
-            name: PRE-RELEASE DEPENDENCIES
-
-    name: ${{ matrix.name }} Python ${{ matrix.python }}
-
-    env:
-      OS: ${{ matrix.os }}
-      PYTHON: ${{ matrix.python }}
+        os: [ubuntu-latest]
+        include: ${{ fromJson(needs.get-environments.outputs.matrix) }}
```
grst, line 54: I don't understand why this doesn't work. The test job in `cookiecutter-scverse-instance` only runs for a single environment even though the json looks exactly as I would expect: 

image

https://github.com/scverse/cookiecutter-scverse-instance/pull/195

{{cookiecutter.project_name}}/.github/workflows/test.yaml
grst: that's a mistake, good catch

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -18,44 +18,61 @@ defaults:
...
+        run: |
+          uvx hatch python install ${{ matrix.python }}
+          uvx hatch env create ${{ matrix.env_name }}
       - name: run tests using hatch
         env:
           MPLBACKEND: agg
           PLATFORM: ${{ matrix.os }}
           DISPLAY: :42
-        run: uvx hatch test --cover --python ${{ matrix.python }}
+        run: uvx hatch run ${{ matrix.env_name }}:run-cov
```
grst, line 73: Apparently this requires to manually create the environment before, while `uvx hatch test` does not.

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -18,44 +18,61 @@ defaults:
...
-          filter: blob:none
-          fetch-depth: 0
       - name: Install uv
         uses: astral-sh/setup-uv@v5
         with:
           cache-dependency-glob: pyproject.toml
+      - name: create hatch environment
+        run: |
+          uvx hatch python install ${{ matrix.python }}
+          uvx hatch env create ${{ matrix.env_name }}
```
grst, line 67: Apparently this requires to setup the correct python version before, while directly using `hatch test` does not. 
Additionally, it still fails with 

```
 Installing 3.8
Installed 3.8 @ /home/runner/.local/share/hatch/pythons/3.8

The following directory has been added to your PATH (pending a shell restart):

/home/runner/.local/share/hatch/pythons/3.8/python/bin
Environment `hatch-test.py3.8` is incompatible: cannot locate Python: 3.8
```

which I don't get.

grst (commented):
Hi @flying-sheep, 

there are still some issues that need to be resolved, maybe you have an idea for some of them. 

Overall, I'm also still a bit hesitant if this is the way to go. The `jq` command is a bit obsure to someone who hasn't worked with that before, and I feel it makes it very hard for developers using the template to understand what's actually going on... whereas

```
         - os: ubuntu-latest
            python: "3.12"
          - os: ubuntu-latest
            python: "3.12"
            pip-flags: "--pre"
            name: PRE-RELEASE DEPENDENCIES
```

is pretty explicit.

flying-sheep:
AH, missed that today. I’ll take a look on monday!

{{cookiecutter.project_name}}/.github/workflows/test.yaml
flying-sheep: Only thing I can see is that you typed `fromJson` instead of `fromJSON`

{{cookiecutter.project_name}}/.github/workflows/test.yaml
flying-sheep: Another option is 

```yaml
- uses: astral-sh/setup-uv@v5
  with:
    python-version: ${{ matrix.python-version }}
```

grst: > pending a shell restart

that's it probably. 

But passing it to `setup-uv` is probably the best way forward as you suggested.

flying-sheep:
If we get this to work, we should probably also use https://github.com/marketplace/actions/alls-green#why

{{cookiecutter.project_name}}/.github/workflows/test.yaml
flying-sheep: Ah, the issue is the overcomplicated way `include` works: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations

flying-sheep: I think in order to do what we want, we need to define our matrix using both `os` and `env_name` arrays, and then “[expand](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-expanding-configurations)” the configurations based on `env_name`.

```yaml
matrix:
  os: [ubuntu-latest]
  env_name: [...]
  include:
  - env_name: ...
    name: ...
    python: ...
  ...
```

flying-sheep:
So for excluding the incompatible python versions, the only thing I can think of is manually excluding them because of https://github.com/pypa/hatch/issues/1571

When you run `hatch test -a`, hatch will skip the incompatible ones, but of course we should avoid creating jobs for incompatible Python versions.

flying-sheep:
Good point! I checked how it would look like to use a Python script instead, and it’s pretty OK, but this is more local.

6135fc4901376f02a183e5d5d9f94f0c90514aa7

flying-sheep:
> The jq command is a bit obsure to someone who hasn't worked with that before, and I feel it makes it very hard for developers using the template to understand what's actually going on...

How big do you think the overlap is between “needs to customize the test jobs in a way that isn’t covered by editing pyproject.toml” and “doesn’t know jq”? Not a leading question, I don’t know the answer here.

I think combined with https://github.com/marketplace/actions/alls-green#why, this PR would make it so few people need to touch this workflow, whereas with a duplicated matrix, people do have to sync them.

flying-sheep:
I think this is ready now, the failure is unrelated: https://github.com/scverse/cookiecutter-scverse-instance/pull/195#issuecomment-2812253121

flying-sheep (approved):

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -48,14 +63,28 @@ jobs:
...
         uses: codecov/codecov-action@v4
+
+  check: # https://github.com/re-actors/alls-green#why
+    if: always()
+    needs:
+      - get-environments
+      - test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: re-actors/alls-green@release/v1
```
grst, line 88: I suggest to use a more descriptive name, currently it just shows up as "check" 

image


```suggestion
      - uses: re-actors/alls-green@release/v1
        name: Tests pass in all hatch environments
```

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -48,14 +63,28 @@ jobs:
...
           cache-dependency-glob: pyproject.toml
+      - name: create hatch environment
+        run: uvx hatch env create ${{ matrix.env.name }}
       - name: run tests using hatch
         env:
           MPLBACKEND: agg
           PLATFORM: ${{ matrix.os }}
           DISPLAY: :42
-        run: uvx hatch test --cover --python ${{ matrix.python }}
+        run: uvx hatch run ${{ matrix.env.name }}:run-cov
```
grst, line 75: Just to be sure the command names `run-cov` and `coverage` can be considered stable in hatch?

{{cookiecutter.project_name}}/.github/workflows/test.yaml
```diff
@@ -18,27 +18,42 @@ defaults:
     shell: bash -euo pipefail {0}
 
 jobs:
+  get-environments:
```
grst, line 21: How about adding some comments on what this is doing and why we think this is the preferred way?

grst (commented):
Pretty neat, thanks for fixing the github actions! I think we could improve a bit on explaining what this is all about, but apart from that I think this is good to go. 

(And I decided that the advantage of having the `pyproject.toml` as single point of truth outweighs the additional complexity)

{{cookiecutter.project_name}}/.github/workflows/test.yaml
flying-sheep: Hmm, one could also add other jobs to it, but I guess then people can also change the name

{{cookiecutter.project_name}}/.github/workflows/test.yaml
flying-sheep: Sounds good!

PR #394, opened 2025-04-18
Re-enable dry-run for release

grst:


grst:
I want to test again that everything works as expected in dry-run mode befor actually triggering the next template sync.

PR #400, opened 2025-05-08
Switch to uv for RTD build

### LINKED ISSUE #399: Use UV for RTD build
https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.readthedocs.yaml

```python
version: 2
build:
    os: ubuntu-24.04
    tools:
        python: "3.13"
    commands:
        - asdf plugin add uv
        - asdf install uv latest
        - asdf global uv latest
        - uv venv
        - uv pip install .
        - .venv/bin/python -m sphinx -T -b html -d docs/_build/doctrees -D language=en docs $READTHEDOCS_OUTPUT/html
sphinx:
    configuration: docs/conf.py
    fail_on_warning: false

submodules:
    include: all
```

there's a less verbose way by just installing `uv` but this is the more official way

Zethson:
Fixes https://github.com/scverse/cookiecutter-scverse/issues/399

- Use UV for RTD build which can substantially speed up doc generation & preview


grst:
do you by any chance know if there's also a `hatch` plugin for readthedocs? This would tie in well with @flying-sheep and my efforts to standardize builds between local machines and CI. 

We could also execute hatch through uv I suppose.

flying-sheep (changes requested):
`fail_on_warning` *must* be true by default, and the comment should read “disable this if you are OK with broken links in your docs”. I’m losing my mind here, why does everybody just love to try and change this every single time this file gets touched?

In scanpy I needed to switch it to this before people kept their ill-advised fingers off it.

```yaml
fail_on_warning: true # do not change or you will be fired
```

Besides: does it even do anything if we call `sphinx` manually above? Is there a SPHINX_BUILD_OPTS envvar that gets set by this configuration or something? We should use that or get rid of the whole `sphinx:` block if it does nothing.

and add `-W` to the command line of course.

Zethson:
`fail_on_warning must be true by default, and the comment should read “disable this if you are OK with broken links in your docs”. I’m losing my mind here, why does everybody just love to try and change this every single time this file gets touched?`

That's just copy and paste. We'll stick to true :)

flying-sheep:
phew! this question remains though:

> does [the `sphinx:` config block] even do anything if we call sphinx manually above? Is there a SPHINX_BUILD_OPTS envvar that gets set by this configuration or something? We should use that or get rid of the whole `sphinx:` block if it does nothing.

Zethson:
I don't think that it still does anything. You are correct.

I added `-W` and removed the section.

Reference: https://docs.readthedocs.com/platform/stable/build-customization.html#install-dependencies-with-uv

Zethson:
Closing in favor of #402

PR #402, opened 2025-05-08
Use hatch on RTD

grst:
Trying to get RTD to build the docs with hatch. 
If this works, it would be an alternative to #400

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -61,7 +61,7 @@ jobs:
       - name: Build the documentation
         run: |
           cd "$PROJECT_ROOT"
-          hatch run docs:build -W
+          hatch run docs:build
 
   test-scripts:
     runs-on: ubuntu-latest
--- a/{{cookiecutter.project_name}}/.readthedocs.yaml
+++ b/{{cookiecutter.project_name}}/.readthedocs.yaml
@@ -1,16 +1,14 @@
 # https://docs.readthedocs.io/en/stable/config-file/v2.html
 version: 2
 build:
-  os: ubuntu-20.04
-  tools:
-    python: "3.10"
-sphinx:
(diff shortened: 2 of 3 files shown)


grst:
Down from 44s to 26s.

Zethson (commented):
Great!

However, I wonder whether we lose the fail on warning here.

https://github.com/scverse/cookiecutter-scverse/blob/1bf0b5448a2e0f28c6c902024207e3e02755393a/%7B%7Bcookiecutter.project_name%7D%7D/pyproject.toml#L71 does not have a `-W` and we're deleting the sphinx section.

grst:
it should probably be added to `pyproject.toml` then.

Zethson (approved):
Now I think it should be good.

grst:
Not sure why we didn't have this in `pyproject.toml` until now? if we think it's desirable to *only* have this enabled on RTD by default, the `-W` flag could also be appended to the `hatch run` command.

flying-sheep:
Read the last sentence of your link:
> If you are using the `build.commands` feature, you don't need to make any changes, as the `sphinx` and `mkdocs` options are not required when using `build.commands`.

flying-sheep (approved):
About the `-W` flag: [since Sphinx 8.1](https://www.sphinx-doc.org/en/master/changes/8.1.html#features-added) it no longer cancels the build, so we get a `-1` exit code and also fully built docs, which I think is perfect for the CLI. “Something is wrong” but you can inspect the complete docs.

I wish it was possible to have RTD’s PR status check fail while also having it upload the docs, that would be perfection.

I updated the dependency to `sphinx >=8.1` so users can rely on that behavior.

grst:
I think if you provide a custom build command, the entire `sphinx:` context is ignored

PR #406, opened 2025-05-15
Update documentation

### LINKED ISSUE #330: more background on hatch
Opening an issue with the info from #328 :

@jan-engelmann
> I think the docs would greatly benefit of a little introduction section on hatch and what needs to be done such that hatch python and not conda is used. I know that this is hatch and not cookiecutter specific but I think it would help adoption.

@eroell 
I agree that hatch could for new users be explained in a few more words.

grst:
- [ ] in template_usage, several markdown links are broken
 - [ ] update template sync section
 - [ ] update info on hatch, incl. adding more background. 
 - [ ] show how to use pip or uv with hatch

Close #330

### DIFF
--- a/{{cookiecutter.project_name}}/docs/template_usage.md
+++ b/{{cookiecutter.project_name}}/docs/template_usage.md
@@ -171,7 +171,6 @@ On the RTD dashboard choose "Import a Project" and follow the instructions to ad
   This ensures that a PR doesn't introduce changes that break the documentation.
   To do so, got to `Admin -> Advanced Settings`, check the `Build pull requests for this projects` option, and click `Save`.
   For more information, please refer to the [official RTD documentation][rtd-prs].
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_name}}/docs/contributing.md
```diff
@@ -1,44 +1,107 @@
...
+```
+
+::::
+
+::::{group-tab} uv
+
+A popular choice for managing virtual environments is [uv][].
+The main disadvantage compared to hatch is that it supports only a single environment per project at a time,
+which requires you to mix the dependencies for running tests and building docs.
+This can have undesired side-effects,
```
flying-sheep, line 72: yeah, that’s why I wouldn’t go with your suggestion in https://github.com/scverse/cookiecutter-scverse/issues/424#issuecomment-3265441645

PR #426, opened 2025-09-09
Simplify hatch envs

grst:
@flying-sheep, given our discussion in #406, I believe the way to go would then simply be to include `dev` dependencies in the test-env and tell users to use the test environment as default in vscode. 

WDYT?


flying-sheep:
That’s how I do it for the projects I maintain, so I think it makes sense!

PR #430, opened 2025-09-25
Enable parallel testing with hatch

grst:
See also https://github.com/pypa/hatch/issues/2048


flying-sheep:
Oooh, `coverage.py` actually recently improved this! Since 7.10.0 (released in July), there’s [`[run] patch`](https://coverage.readthedocs.io/en/latest/config.html#run-patch) which sets all of this up!

PR #443, opened 2025-11-14
feat: runnable bug reports

ilan-gold:


ilan-gold:
I would be open to making it optional, but strongly encouraged.

ilan-gold:
>If someone doesn't bother to make a reproducible example already now, they'll also leave the new field empty if it's optional.

I will say, the motivation was that we got a bug report that was "incomplete" and my feeling was that if we made this process easier, we would have gotten a complete one (and then could have fixed the issue in 1 day instead of waiting 2 months to get enough repro info).

flying-sheep:
> we got a bug report that was "incomplete"

we constantly get a lot of them, and most are a waste of time since we try to reproduce and then just can’t.

I think whatever we can to do inspire people to post more complete examples, that would be ideal

Zethson:
Since you've already applied this change to Anndata and/or scanpy, I suggest that we observe what your experiences are before merging this change.
Do you see an uptick in reproducible issues or are you suddenly getting a lot less issues? Let's find out. 

Wdyt?

ilan-gold:
I'm fine waiting.  I'm glad that we started the discussion then :) Let's see!

PR #446, opened 2025-11-20
chore: use KaTeX Sphinx plugin

flying-sheep:
KaTeX renders to static HTML and doesn’t re-layout the page a fraction of a second after load.

compare (you might have to use devtools to see it if you have the page in cache):

mathjax: https://docs.scvi-tools.org/en/stable/api/reference/scvi.model.base.DifferentialComputation.html#scvi.model.base.DifferentialComputation.get_bayes_factors

katex-pre-render: https://cookiecutter-scverse-instance--237.org.readthedocs.build/en/237/generated/cookiecutter_scverse_instance.pp.elaborate_example.html


grst (approved):
Just one minor concern that it's one additional thing that can go wrong while building docs. But it's probably fine.

PR #457, opened 2025-12-23
Enable OIDC in Codecov action

### LINKED ISSUE #454: codecov.io now wants a token
the coverage job should therefore look like [this](https://github.com/scverse/scanpy/blob/452c4f6915a1315f55c91aa174e6b19259046801/.github/workflows/ci.yml#L76-L82):

```yaml
- name: Upload coverage data
  if: ${{ !cancelled() }}
  uses: codecov/codecov-action@v5
  with:
    token: ${{ secrets.CODECOV_TOKEN }}
    fail_ci_if_error: true
```

- `if: ${{ !cancelled() }}` (maybe) to upload coverage even when tests fail.
- `fail_ci_if_error: true` to prevent that step from silently not working

and we should mention to (how to) add the CODECOV_TOKEN secret if we don’t already.

grst:
- **use codecov with OIDC**
- **Update Codecov docs**

Close #454


flying-sheep (approved):
Beautiful! Did you check that it works?

grst:
On the cookiecutter-scverse-instance PR it worked. It's not from a fork though. I can PR the instance repo from a fork after this is merged.

PR #461, opened 2025-12-29
fix: typo

ilan-gold:


ilan-gold:
How do you think would be best to change the github links on the website to point at this repo instead of the instance? That's what brought me to the other repo. Would updating this break things?

https://github.com/scverse/cookiecutter-scverse/blob/6a1f5a101d3de6ef0c48821eaaa00d078ad223a1/%7B%7Bcookiecutter.project_name%7D%7D/pyproject.toml#L19

grst:
It would have to be made conditional on the `cookiecutter._render_devdocs` variable, then it will only affect the instance repo. As it's done e.g. here: 

https://github.com/scverse/cookiecutter-scverse/blob/6ff5b92b5d44ea6d8a88e47538475718d467db95/%7B%7Bcookiecutter.project_name%7D%7D/docs/index.md?plain=1#L11-L13

Issue #113, opened 2022-10-24
Extract custom sphinx stuff to package


flying-sheep:
#100 did this:

> - Remove scanpydoc dependency
> - Add typed returns custom extension like in scanpy
> - Use same autosummary class template as in scvi-tools (single page for classes with methods/attr tables)

why? shared code should live in a package. let’s revert that part and improve scanpydoc to enjoy shared goodness.

grst:
Is it that bad to have a bit of vendored code, especially with template
sync in place?

adamgayoso:
> Is it that bad to have a bit of vendored code, especially with template
sync in place?

This is also a good point. I'm indifferent. Package makes it easier for people to use without using the template.

flying-sheep:
I believe minimizing boilerplate and relying just on Git (for your project specific settings) and dependencies (for things maintained by other people) simplifies things. Sure, after learning Git and Python dependency management, some people also want to learn [cruft](https://github.com/cruft/cruft) and configure it to deal with boilerplate, but how many of our users are that?

What do you think? How realistic is it that our users will use template sync?

grst:
@flying-sheep, I think the only custom sphinx stuff we are still using is [typed returns](https://github.com/scverse/cookiecutter-scverse/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/docs/extensions/typed_returns.py). Do you know if we still need this or if it maybe got implemented in autodoc-typehints by now?

grst:
We are now using a custom sphinx extension from scverse-misc. There's also a separate discussion on harmonizing the docs experience across scverse packages. While this will eventually find its way into the cookiecutter template, it's a separate discussion, therefore, I'm closing this now.

flying-sheep:
Also https://github.com/scverse/scverse-doc

PR #483, opened 2026-03-02
Docs fixes / improvements

ilia-kats:
Various fixes for the autosummary templates. In particular:

- don't print Methods table heading if there are no methods

  Currently, the methods heading is printed if there are any methods, including `__init__`, but `__init__` is filtered out from the methods list and table. This can lead to the situation that if a class has only `__init__` and no other public methods, the "Methods table" and "Methods"  headings will be printed, but no actual methods will be documented. Move the methods filtering to the very beginning and overwrite the methods variable with the filtered version to prevent above problem.

- use fullname instead of objname in the class template

  With this change, one can apply autosummary to submodules of the package and have everything work.

- add a module.rst autosummary template

  The only difference to the default template is the addition of toctree directives for functions, classes, and exceptions. This enables one to pass a module to autosummary and have it generate documenation pages and crossreferences for all top-level functions, classes, and exceptions, not just for submodules.


grst:
I'm not a huge fan of adding yet another custom file to the sphinx configuration. 
We earlier discussed making the docs customizations a package https://github.com/scverse/cookiecutter-scverse/issues/113

Could this be part of `scverse-misc[docs]`? 
Is `scanpydoc` still actively used? Can it be replaced by scverse-misc? 
Or is it better to have a separate package (e.g. `scversedocs`)?

ilia-kats:
If you're talking about the new extension, that's a workaround for https://github.com/sphinx-doc/sphinx/issues/14264. I'm happy to also put it into scverse-misc (might go well with https://github.com/scverse/scverse-misc/pull/37), scanpydoc, or wherever, but then the cookiecutter template should use that package by default. Also note that that extension is tightly integrated with the `class.rst` template, so if we move it, we need to move everything.

grst:
> but then the cookiecutter template should use that package by default

agree, if everything that is now a custom sphinx configuration in this cookiecutter template is covered by scverse-misc, I'm happy to depend on it by default.

PR #485, opened 2026-03-12
support prek in addition to pre-commit

ilia-kats:
The `post_gen_project` hook currently fails if `pre-commit` is not installed. This PR adds additional support for `prek`.


flying-sheep:
neat, thanks! I think that should be all that’s necessary, and if not, we can always fix it

PR #487, opened 2026-03-30
Take up template sync where it got interrupted

### LINKED ISSUE #459: If update was successful within forked repo, but PR was not, then no PR will be created when rerunning
During v0.7 release we had an issue with wrong token permissions that allowed to fork and update the repo, but PR creation failed. 
When fixing the permissions, and rerunning the sync workflow, the repos that ran through in the previous run did not get PRed. 

The reason is, that we check if the repo has been updated before creating the PR: 
https://github.com/scverse/cookiecutter-scverse/blob/6ff5b92b5d44ea6d8a88e47538475718d467db95/scripts/src/scverse_template_scripts/cruft_prs.py#L462

grst:
Close #459


.github/workflows/cruft-prs.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         run: uvx --from ./scripts send-cruft-prs ${{ env.RELEASE }} --all_repos --log-dir log
         env:
           RELEASE: ${{ github.event_name == 'release' && github.event.release.tag_name || github.event.inputs.release }}
-          GITHUB_TOKEN: ${{ secrets.BOT_GH_TOKEN }}
```
grst, line 28: moving away from using an org-wide secret. See notion page for more details.

grst:
Docs failure is unrelated. See also #508, but I currently hope it's just a temporary issue.

scripts/src/scverse_template_scripts/cruft_prs.py
```diff
@@ -514,24 +514,27 @@ def make_pr(con: GitHubConnection, release: GHRelease, repo_url: str, *, log_dir
     if dry_run:
         log.info("Skipping PR because in dry-run mode")
         return
-    if updated:
```
flying-sheep, line 517: why no longer use this bool? is there no realistic chance that it’s `False`?

flying-sheep (approved):

scripts/src/scverse_template_scripts/cruft_prs.py
grst: Using that bool in the first place is part of the problem I'm trying to address ;) 

When checking for `if updated`, the PR will only be created if a commit was made to the forked repo. 
However, it can be (and this happend in the past) that the process gets interrupted after making the commit and before making the PR. If you then trigger a rerun of the pipeline, no commit will be created (because it's already there) and making the PR will be skipped although it doesn't exist yet. 

If we don't perform this check, the pipeline simply picks up where it left off after an interruption.

flying-sheep: got it!

PR #488, opened 2026-03-30
Switch to github app for PR creation

grst:


.github/workflows/deploy-instance-repo.yml
```diff
@@ -16,30 +16,40 @@ jobs:
...
-        uses: actions/checkout@v6
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          cache-dependency-glob: .github/workflows/deploy-instance-repo.yml
 
       - name: Set git identity
         run: |
           git config --global user.name "scverse-bot"
           git config --global user.email "108668866+scverse-bot@users.noreply.github.com"
```
flying-sheep: this should changed to this, right?

- name: `scverse-pr-creator[bot]`
- email: `272303624+scverse-pr-creator[bot]@users.noreply.github.com`

I got the ID via

```console
$ gh api /users/scverse-pr-creator[bot] --jq .id
272303624
```

.github/workflows/deploy-instance-repo.yml
flying-sheep: I see, this is why these commits aren’t attributed to the bot (now app).

What uses the credentials set by `git config --global` then?

grst: I think nothing really uses them, it's just that the calls to the `git` CLI within the action fail if no user is set up.

flying-sheep (approved):

PR #495, opened 2026-04-17
feat: remove __version__ cargo cult

flying-sheep:
`__version__` is an old magic spell that has been cargo-culted to modern times without need. Some people even derived more magic from it like `__author__`, but nothing caught on so tenaciously as `__version__`.

Much breath has been wasted over purported health benefits such as “performance”. I won’t repeat all that here, just trust me: we shouldn’t propagate this vestige.

People can and should just use `importlib.metadata.version('pkgname')`.

> [There should be one– and preferably only one –obvious way to do it.](https://peps.python.org/pep-0020/)

## compatibiliy

this is backwards-compatible, as `__version__` is neither in the docs nor in `__all__` and therefore not a public API.

### DIFF
--- a/{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/__init__.py
+++ b/{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/__init__.py
@@ -1,7 +1,3 @@
-from importlib.metadata import version
-
 from . import pl, pp, tl
 
 __all__ = ["pl", "pp", "tl"]
-
-__version__ = version("{{ cookiecutter.project_name }}")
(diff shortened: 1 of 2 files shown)


grst:
While I agree that people should use importlib, there's still a lot of functionality built around this convention. 
I don't think it's wrong providing backwards-compatibility with this approach for such a little cost.

flying-sheep:
There’s only `session-info` AFAIK, which not only is very old, but also very tenuously maintained (don’t look at its code :fearful:). I built `session-info2` to replace it and it seems to hold up nicely, there has been exactly one bug so far (around `.pth` handling, was fixed quickly).

grst (approved):

grst:
I thought about it again and came to the conclusion that you are right. If at this point anyone relies on __version__ it's a bug on their end.

PR #498, opened 2026-04-28
chore: zizmor for pre-commit

ilan-gold:


grst:
@ilan-gold do we still want this or did you find a different solution?

ilan-gold:
I think we landed on "let's see how it goes in anndata" but I haven't merged it there yet.

ilan-gold:
So this seems to be going well - dependabot does indeed understand the commit-hash/version syntax this would impose and does the updates e.g., https://github.com/scverse/anndata/pull/2482

@flying-sheep was there anything else we were concerned about?

grst:
This makes only sense in combination with dependabot, I guess? The standard pre-commit autoupdates won't work anymore?

ilia-kats:
Why wouldn't the pre-commit autoupdates work? This is completely orthogonal: pre-commit updates `.pre-commit.yaml`, dependabot updates the GitHub actions. But yes, I would include a dependabot config in this PR and also fix the shipped workflows to pass zizmor's audits.

grst:
ah, I had assumed it would also pin hashes in .pre-commit-config.yaml

ilia-kats:
zizmor is specifically for GH actions. pre-commit itself supports pinning hashes in `.pre-commit.yaml`, and the CLI can do `autoupdate --freeze` to update the hashes, but I can't find anything on what the GH app does.

grst:
According to https://github.com/pre-commit/pre-commit/issues/3488 it "already supports the freeze format and maintains it in place". But would be good to verify. 

Or maybe even better: switch to dependabot for updating pre-commit configs? 
https://github.blog/changelog/2026-03-10-dependabot-now-supports-pre-commit-hooks/

ilia-kats:
> According to [pre-commit/pre-commit#3488](https://github.com/pre-commit/pre-commit/issues/3488) it "already supports the freeze format and maintains it in place". But would be good to verify.

I've switched [mofaflex](https://github.com/biofam/mofaflex) to using hashes, I guess we'll see next Monday if that works.

> Or maybe even better: switch to dependabot for updating pre-commit configs? https://github.blog/changelog/2026-03-10-dependabot-now-supports-pre-commit-hooks/

The app is still useful for checking / fixing pull requests etc., and I don't see an [option](https://pre-commit.ci/#configuration) to turn off autoupdates for the app.

grst:
> The app is still useful for checking / fixing pull requests etc., and I don't see an [option](https://pre-commit.ci/#configuration) to turn off autoupdates for the app.

ok, would be an option then if we decide to switch to pre-commit CI lite, or autofix.ci

.github/workflows/cruft-prs.yml
```diff
@@ -12,23 +12,26 @@ jobs:
...
       - name: Set git identity
         run: |
           git config --global user.name "scverse-bot"
           git config --global user.email "108668866+scverse-bot@users.noreply.github.com"
       - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v7
+        uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
         with:
           cache-dependency-glob: scripts/pyproject.toml
+          enable-cache: false
```
grst, line 28: @ilia-kats do we really want this? But I guess the speed impact is actually minor with `uv`.

ilia-kats: I've read the [article](https://adnanthekhan.com/2024/12/21/cacheract-the-monster-in-your-build-cache/) that zizmor's documentation links to for this issue (highly recommended), and from a security standpoint it would be good to disable cache for this. Briefly, malicious PRs can poison the cache to inject arbitrary code into the package, so if my understanding is right, the script that creates the PRs to the template repos could be overwritten with something that injects malicious code in the PRs, for example.

Speed impact should be minimal with uv, but even if it's not, I think the bottleneck in this workflow is the PR creation, not setup.

.github/workflows/test.yml
```diff
@@ -6,9 +6,13 @@ on:
   pull_request:
     branches: [main]
 
+permissions: {}
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    permissions:
```
ilia-kats: Both jobs in this file need the same permissions, so might as well put this into the global permissions block. Also for all other workflows, my persional preference is to put `contents: read` into the global permisisons block: Almost everything needs to be able to read the repo to do anything useful, and you're still overriding the default overly broad permssions, plus you don't need to remember to add this to every job in the file.. But as I said, just a personal preference, feel free to ignore.

ilia-kats (approved):
Looks good to me, with one minor nitpick that you can ignore if you want.

.github/workflows/test.yml
grst: sounds good!
I left `permissions: {}` in the some of the outer (i.e. template managing) github actions that really don't need this permission as they only do API calls. 

I also left it for the release action, as it's quite short and explicitly requests `contents: read` and `id-token: write` permissions within the job.

.github/workflows/propagate-pre-commit.yml
```diff
@@ -5,22 +5,28 @@ on:
...
 
+permissions: {}
+
 jobs:
   update:
     runs-on: ubuntu-latest
-    if: ${{ github.actor == 'pre-commit-ci[bot]' }}
+    permissions:
+      contents: write # for stefanzweifel/git-auto-commit-action
+    if: ${{ github.actor == 'pre-commit-ci[bot]' }} # zizmor: ignore[bot-conditions] -- action does not run in privileged context (e.g. pull_request_target) and does not execute any user-provided code
```
ilia-kats, line 15: The zizmor docs recommend using `github.event.pull_request.user.login` here. `github.actor` refers to the person who last modified a PR (and can be triggered if the bot pushed to a malicious PR for example), while `github.event.pull_request.user.login` is the actual creator of the PR. The value should be identical (`pre-commit-ci[bot]`).

Even if this is harmless, probably still better to change it rather than ignore. Apologies for missing this before.

ilia-kats: Or even better, using `github.event.pull_request.user.id == 66853113` (according to [the API](https://api.github.com/users/pre-commit-ci[bot]))

grst: I've read this, but this triggers on a push event, so not sure if `pull_request.user` is available here? But it could be, since it's still within a PR.

ilia-kats: Ah sorry, missed that this is on a push event. Then it's fine, if some has push access to this repo there is nothing more to secure.

PR #508, opened 2026-06-08
Remove uv installation from readthedocs config.

grst:
It's now available natively.

https://about.readthedocs.com/blog/2026/04/uv-native-support/


grst:
Does not work as I expected

PR #511, opened 2026-06-08
Allow user to choose between issue type and labels

### LINKED ISSUE #492: Issue template may be broken
https://github.com/lucas-diedrich/mulink/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml is broken for a reason I don't fully get, but seems to have something to do with org-level permissions around issues judging [by github's docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax):

```
The issue type that will be automatically added to issues created with this template.
Issue types are defined at the organization level and can be used to create a shared syntax across repos.
```

I don't see anything in https://cookiecutter-scverse-instance.readthedocs.io/en/latest/template_usage.html about this and also have never had this issue myself.  Is there something we can add to the documentation about this?

cc @lucas-diedrich

grst:
- **Let users choose between issue labels and issue types (#492)**
- **update question text**

Closes #492

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -15,6 +15,10 @@
         "Unlicense"
     ],
     "ide_integration": true,
+    "issue_categorization": [
+        "labels",
+        "issue types"
+    ],
     "_copy_without_render": [
         ".github/workflows/build.yaml",
         ".github/workflows/test.yaml",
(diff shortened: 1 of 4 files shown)


cookiecutter.json
```diff
@@ -45,6 +46,7 @@
         "author_email": "The e-mail address your package’s users can contact you under",
         "github_user": "The GitHub username or org the project is to be published under",
         "github_repo": "If the repo name should differ from the project name, edit it now",
-        "ide_integration": "Whether to generate IDE configuration files"
+        "ide_integration": "Whether to generate IDE configuration files",
+        "issue_categorization": "How to categorize issues in the issue templates. “issue types” only works in GitHub organizations, while  “labels” also works in personal repos."
```
flying-sheep, line 50: too bad we have no response in https://github.com/orgs/community/discussions/175785 if this is intentional or just temporary.

flying-sheep (approved):

grst:
Ok, I'm going to merge this for now. Should they ever implement this feature we can roll back the changes.

PR #513, opened 2026-06-15
README: use the official readthedocs badge

ilia-kats:
the shields.io badge started displaying "rate-limited by upstream service" most of the time

### DIFF
--- a/{{cookiecutter.project_name}}/README.md
+++ b/{{cookiecutter.project_name}}/README.md
@@ -4,7 +4,7 @@
 [![Documentation][badge-docs]][documentation]
 
 [badge-tests]: https://img.shields.io/github/actions/workflow/status/{{ cookiecutter.github_user }}/{{ cookiecutter.project_name }}/test.yaml?branch=main
-[badge-docs]: https://img.shields.io/readthedocs/{{ cookiecutter.project_name }}
+[badge-docs]: https://app.readthedocs.org/projects/{{ cookiecutter.project_name }}/badge/?version=latest
 
 {{ cookiecutter.project_description }}
 


grst (approved):

{{cookiecutter.project_name}}/README.md
```diff
@@ -4,7 +4,7 @@
 [![Documentation][badge-docs]][documentation]
 
 [badge-tests]: https://img.shields.io/github/actions/workflow/status/{{ cookiecutter.github_user }}/{{ cookiecutter.project_name }}/test.yaml?branch=main
-[badge-docs]: https://img.shields.io/readthedocs/{{ cookiecutter.project_name }}
+[badge-docs]: https://app.readthedocs.org/projects/{{ cookiecutter.project_name }}/badge/?version=latest
```
flying-sheep, line 7: don’t hardcode `version`, people can choose which one is the default.

That’s why links to HTML pages should optimally contain `/page/` too instead of `/en/latest/` or so

ilia-kats: I was under the impression that this corresponds to what shields.io is doing?

flying-sheep: On RTD, you choose a “default version”, which is the one referenced when you use `/` or `/page/<subpage>` links. And that’s also the one you get when not specifying `version=` for the badge.

I doubt shields.io uses anything other than the default version when you don’t specify one, since `latest` is just one of the versions that are auto-created for a new project, and it can be deleted or renamed.

/edit: yup, it just forwards to the RTD badge, which of course defaults to the default version: https://github.com/badges/shields/blob/dffd223ef4c0026c786c8e0d5cd93f9cc24bec55/services/readthedocs/readthedocs.service.js#L60-L66

PR #516, opened 2026-06-19
WIP agent-based sync

### LINKED ISSUE #509: agent-based sync
The current template sync can lead to a ton of merge conflicts that are tedious to resolve. 
We could explore replacing it by an agent "skill". 

In that case we could either 
 * centrally make PRs (will cost us tokens) or
 * simply open issues on ecosystem packages whenever a new version is available and refer to corresponding instructions. Developers could then assign the issue to an agent on their end.

grst:
Close #509 

fully vibe-coded prototype, just for further discussion if this makes sense and how this could look like.


Zethson:
I wonder whether this can also cover the "I have an existing non-scverse repository and want to move it to the scverse cookiecutter". We might need a separate skill for that but there'll likely be some overlap. I'm happy to try this PR out when it's ready.

PR #524, opened 2026-07-06
Recommend uv for installation

### LINKED ISSUE #523: Recommend using `uv` instead of `pip`
In the readme we see `pip install git+https://github.com/{{ cookiecutter.github_user }}/{{ cookiecutter.github_repo }}.git@main`. Shall we replace this to an `uv` command?

CC @ajkswamy

grst:
Close #523

### DIFF
--- a/{{cookiecutter.project_name}}/README.md
+++ b/{{cookiecutter.project_name}}/README.md
@@ -21,18 +21,39 @@ If you don't have Python installed, we recommend installing [uv][].
 There are several alternative options to install {{ cookiecutter.project_name }}:
 
 <!--
-1) Install the latest release of `{{ cookiecutter.project_name }}` from [PyPI][]:
+1. Install the latest release of `{{ cookiecutter.project_name }}` from [PyPI][]:
+
+We recommend using [uv][] to add `{{ cookiecutter.project_name }}` to your project's virtual environment:
+
+```bash
+uv add {{ cookiecutter.project_name }}
+```
+
+<details>
+<summary>If you prefer using `pip`:</summary>
 
 ```bash
 pip install {{ cookiecutter.project_name }}
 ```
+
+</details>
+
 -->
 
 1. Install the latest development version:
 
 ```bash
+uv add git+https://github.com/{{ cookiecutter.github_user }}/{{ cookiecutter.github_repo }}.git@main
+```
+
+<details>
+<summary>If you prefer using `pip`:</summary>
+
+```bash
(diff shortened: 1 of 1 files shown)


{{cookiecutter.project_name}}/README.md
```diff
@@ -21,18 +21,39 @@ If you don't have Python installed, we recommend installing [uv][].
...
+   ```bash
+   uv add git+https://github.com/{{ cookiecutter.github_user }}/{{ cookiecutter.github_repo }}.git@main
+   ```
+
+   <details>
+   <summary>If you prefer using `pip`:</summary>
+
+   ```bash
+   pip install git+https://github.com/{{ cookiecutter.github_user }}/{{ cookiecutter.github_repo }}.git@main
+   ```
```
flying-sheep, line 54: I think `@main` is redundant, why would you want `main` if it’s not the default branch?

Also does the `uv` line work? I know that in some situations uv needs `project-name @ git+https://...`, I’m not sure if `add` is one of these cases. I *think* you need the `project-name @` only for `uv pip` but I’m not sure.

````suggestion
   ```bash
   uv add git+https://github.com/{{ cookiecutter.github_user }}/{{ cookiecutter.github_repo }}.git
   ```

   <details>
   <summary>If you prefer using `pip`:</summary>

   ```bash
   pip install git+https://github.com/{{ cookiecutter.github_user }}/{{ cookiecutter.github_repo }}.git
   ```
````

grst: I tried it, and it works. 

My motivation for `main` was that it shows, as an example, that other git revisions could be specified the same way, but also fine with removing it.

flying-sheep: I see. Hmm, we could add this link in case people want to know more? https://docs.astral.sh/uv/concepts/projects/dependencies/#git

flying-sheep (approved):

Zethson:
I don't have the strongest opinion on this but I still feel that https://github.com/scverse/cookiecutter-scverse/issues/523#issuecomment-4890842109 is generally true. We're adding even more verbosity & tooling complexity here for not thaaaat much gain.

So I'd personally not merge this for now but please go for it if you think otherwise.

grst:
@Zethson, see https://github.com/scverse/cookiecutter-scverse/pull/531 for a compromise that puts emphasis on using venvs rather than using uv. The goal would be to convince people who do not yet use uv or project-specific virtual environments at all to do so.

grst:
Closing in favor of #531

PR #529, opened 2026-07-14
docs: remove custom template and extension, use scverse-misc instead

ilia-kats:
As discussed in #483 .


ilia-kats:
I had to bump the min. Python version for the template to 3.12, because uv requires [all dependency groups to be compatible with `requires_python`](https://docs.astral.sh/uv/concepts/projects/dependencies/#group-requires-python), and `scverse-misc` requires Python 3.12. However, it is only used for docs, so I could also add

```toml
[tool.uv.dependency-groups]
doc = {requires-python = ">=3.12"}
```
to `pyproject.toml` instead.

grst:
np, python 3.11 has anyway already fallen out of SPEC-0. Even 3.12 support will drop soon.

grst (approved):

Issue #504, opened 2026-06-03
bump version to 1.0


grst:
The template is pretty stable now and I only expect incremental changes. Should we bump to v1.0 in the next release? 

Either way, I don't think that "semantic versioning" really applies here, as there's no clearly defined "API" of a template, is there?

Zethson:
There's two things that still bug me majorly that I'll push as time allows:
1. Reviving the type checker discussion. I think that our template has so many advanced topics, it's perfectly fine to have a type checker that mostly only works when code is typed anyways.
2. I'd love for us to already have scversedoc.

But then yeah, we can have a 1.0 release. (Sure, these two things don't block a 1.0 but I want to get them done).

grst:
@flying-sheep any opinion?

flying-sheep:
I agree with Lukas.

One more thing we should decide before 1.0 I think: I wonder if there’s a more general solution than pre-commit.ci for creating a weekly update PR, in which case we could maybe switch to that and either pre-commit.ci lite or [autofix.ci](https://autofix.ci/). I think that’s more flexible as e.g. `mypy` needs all deps duplicated in `extra_dependencies` whereas running custom code would allow to re-use a Python environment for it.

flying-sheep:
Hmm, autofix.ci isn’t super fast, but now I saw an instance of it not committing the fix at all: https://github.com/scverse/anndata/actions/runs/33024507578/job/98362570917?pr=2618

https://github.com/autofix-ci/action/issues/30

flying-sheep:
OK, seems like the reason is that it only runs after the full worflow, and probably  only has access to a short-lived token or so.

So it needs to be set up in its own workflow instead, then it works.

flying-sheep:
wtf, in our terminal output we still link to a document that was renamed in https://github.com/scverse/cookiecutter-scverse/pull/130

4 years ago!

/edit: fix in https://github.com/scverse/cookiecutter-scverse/pull/558

Issue #40, opened 2022-05-05
Come up with a recommended way to render/store tutorials


grst:
I think that wherever it is computationally feasible, tutorials should be run as a CI check and the tutorial use the version rendered by the CI to avoid failing or outdated tutorials. 

With the current template it is already possible to have tutorials built by `nbshpinx` on `readthedocs.org`, but it would be nice to have something that scales better. Also it needs to be documented somewhere how to enable this behavior. 

- [ ] Recommendations where to store tutorials (submodule for executed `ipynb` files when not using CI builds?)
- [ ] Documentation how to enable CI builds
- [ ] Ideally, each notebook is built in a separate worker (-> parallelization) (-> individual checkmark for each notebook instead of digging into sphinx logs) 
- [ ] GPU support

See also #19.

grst:
I'd want something like

```mermaid
flowchart TD
   A[push to PR] --> B[run tests]
   A --> C[build tutorials on GH actions]
   C --> D[trigger RTD build]
```

grst:
@michalk8, can you elaborate how the squidpy setup works? And maybe point me to a PR where the tutorials were actually built, I only find PRs where that particular action was skipped.

grst:
Inspired by the pre-commit CI, the following approach could work: 

1. trigger CI in PR by push
2. rerun notebooks. If the output changes compared to the previous version, add a commit with the rendered ipynb files (either in main repo or submodule)
3. the automatic commit triggers RTD again. Other CI checks can be skipped.

grst:
> Inspired by the pre-commit CI, the following approach could work:
> 
> 1. trigger CI in PR by push
> 2. rerun notebooks. If the output changes compared to the previous version, add a commit with the rendered ipynb files (either in main repo or submodule)
> 3. the automatic commit triggers RTD again. Other CI checks can be skipped.

I think I'll give this approach a try in one of my repos.

grst:
[MyST-nb](https://myst-nb.readthedocs.io/en/latest/quickstart.html) might be a good alternative to nbsphinx. 

Once benefit is that it [supports caching](https://myst-nb.readthedocs.io/en/latest/computation/execute.html) through [jupyter-cache](https://github.com/executablebooks/jupyter-cache). It could be nice to restore that cache when running notebooks through github actions.

flying-sheep:
Isaac wants submodules back in scanpy, see https://github.com/scverse/scanpy/issues/2636

[...]

PR #530, opened 2026-07-15
feat: add mypy type checker and run pre-commit in CI

### LINKED ISSUE #519: Add type checker
I suggest we add a type checker to the pre-commit setup. We had discussed this in the past https://github.com/scverse/cookiecutter-scverse/issues/348#issuecomment-2616035616 and decided against it.

Given the complexity and how opinionated our template is, I think it's totally fine and to add a type checker:

1. Even with a typechecker, typing is optional (given the right configuration). So users that aren't familiar with typing won't be bothered.
2. Typing is incredibly important nowadays and I think it's simple enough to learn. It's not necessary to use advanced typing features or to be super correct about typing.
3. Agents will help solve typing problems and also benefit from typing.
4. If users prefer a different type checker, they can still switch.

There's a few options:

1. **Mypy**: most widely used, often serves as the reference implementation. Mature, but slowest and the error messages can send beginners down rabbit holes (as @flying-sheep noted last time).
2. **Pyright**: fast, excellent inference, but Node-based and the CLI/pre-commit story is less clean. 
3. **ty** (Astral): Extremely fast, and a natural fit since we already standardize on Ruff. Still pre-1.0 and I didn't have the impression it's mature yet.
4. **pyrefly** (Meta): Just recently went 1.0. I've never used it but heard good things.

I think defaulting to mypy would be safest but at the same time, it's not the best option in 2026. I'd like to avoid a node based checker so I'm leaning towards ty or pyrefly.

WDYT?

Zethson:
Closes #519

Adds [mypy](https://mypy.readthedocs.io/) and, as a consequence, switches the template's linting CI off pre-commit.ci onto GitHub Actions + Dependabot.

## mypy

Runs as a **local pre-commit hook that reuses the project environment**, rather than the isolated `mirrors-mypy` hook:

```yaml
- repo: local
  hooks:
    - id: mypy
      name: mypy
      entry: uv run mypy src tests
      language: system
      types: [python]
      files: ^(src|tests)/
      require_serial: true
      pass_filenames: false
```

An isolated hook only sees what's listed in `additional_dependencies`, so you'd have to duplicate (and version-sync) every typed dependency — the pain @flying-sheep raised in #504. Running `uv run mypy` in the project env means mypy checks `src` + `tests` against the **real types of the installed deps** (anndata, numpy, …) with nothing to duplicate, and it scales automatically as a project adds dependencies. (Verified it catches e.g. `"AnnData" has no attribute ...`.)

- `mypy` added to the `dev` group so the hook can resolve it.
- `[tool.mypy] ignore_missing_imports = true` — keeps pytest (in the `test` group) and untyped third-party libs from erroring, while typed deps are still checked for real.
- Typing stays **optional** — mypy only checks annotated functions.

## Off pre-commit.ci → GitHub Actions + Dependabot

The local hook can't run on pre-commit.ci (no project env), which was the nudge to drop pre-commit.ci entirely:

- **`Pre-commit checks` job** in `test.yaml` runs `prek run --all-files` (all hooks, mypy included) and is wired into the existing `check` gate — so everything is enforced in CI without an external service.
- **`.github/dependabot.yml`** (new) keeps hook revs and GitHub Actions versions current (`pre-commit` + `github-actions` ecosystems), replacing pre-commit.ci's autoupdate. Dependabot's `pre-commit` support [landed March 2026](https://github.blog/changelog/2026-03-10-dependabot-now-supports-pre-commit-hooks/) and skips `local` hooks automatically.
- Removed the `ci:` block from `.pre-commit-config.yaml` and the pre-commit.ci setup instructions from the README / `template_usage.md` / `contributing.md`.

## Template code

Only `pp.elaborate_example` failed mypy — it used `TypeVar` placeholders for the optional `mudata`/`spatialdata` packages, which are neither valid types nor runtime-safe (`isinstance` against a `TypeVar` raises). Simplified it to operate on `AnnData`, matching the existing `test_elaborate_example_adata_only_simple` test; `transform` is now `Callable[[np.ndarray], str]`.

## Validation

On a freshly generated project:

- `prek run --all-files` (= the CI `Pre-commit checks` job) — all hooks pass, including mypy over `src` + `tests`
- `hatch run docs:build` (`-W` + `nitpicky`) — build succeeds
- package imports cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -1,14 +1,8 @@
 from collections.abc import Callable, Iterable
-from typing import Any, TypeVar
 
 import numpy as np
 from anndata import AnnData
 
-MuData = TypeVar("MuData")
-SpatialData = TypeVar("SpatialData")
-
-ScverseDataStructures = AnnData | MuData | SpatialData
```
Zethson, line 10: I had introduced this myself but it's a bit complex to make work with mypy when these packages are not installed.
I'll think about it again. Maybe simplifying this again isn't the worst thing either.

Zethson: @flying-sheep actually I'm not really sure how to make this work best without introducing the too heavy dependencies. We have explicit `isinstance` checks and I don't want to change this to `hasattr`. @flying-sheep  do you have any advice or opinion, please?

grst:
in #504, @flying-sheep wrote

> I wonder if there’s a more general solution than pre-commit.ci for creating a weekly update PR, in which case we could maybe switch to that and either pre-commit.ci lite or [autofix.ci](https://autofix.ci/). I think that’s more flexible as e.g. mypy needs all deps duplicated in extra_dependencies whereas running custom code would allow to re-use a Python environment for it.

Do we need to do something about that in the course of this PR? I didn't see you add any `extra_dependnecies` to `.pre-commit-config.yaml` yet, so is this not necessary after all or do you just not use any types from 3rd party libraries?

Zethson:
Ah yeah, thanks.
The current config would indeed not consider any third party stubs. I've now switched mypy to a local config which would reuse the environment. As @flying-sheep pointed out, this would break on pre-commit CI so mypy is skipped on pre-commit CI. Not the biggest fan of that so I wonder whether it'd be best if we stopped recommending people to use pre-commit CI and just provided our own Github Action that'd run the complete pre-commit job.
Advantage: Whole thing would be checked. One less service to take care of.
Disadvantages: pre-commit CI is faster and caches a lot IMO. No autofix but I personally don't find it useful anyways.

grst:
I think it would be nice to consolidate CI services and reduce the amount of services that have to be configured manually. 

We could use
 * dependabot to bump .pre-commit-config.yaml (we'll already use dependabot for #498 and it's enabled by default in github repos)
 * use prek in a custom github action to run pre-commit checks (pre-commit.ci is fast thanks to caching, but prek, unlike pre-commit, uses uv in the background, so it should be quite fast as well).
  * optionally suggest pre-commit.ci lite or autofix.ci, for whoever wants autofixes (the action would still run and flag issues if this is not set up). 
  
Using pre-commit.ci lite or autofix.ci would again tie in nicely with #40 if we eventually get to that.

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
Zethson: Discussed in person. Vendoring type-stubs would be an option but it adds too much complexity. We'll just simplify this to AnnData only.

flying-sheep:
To sum up the conversation between @Zethson and me:

- local mypy plus workflow that runs prek and then commits changes via autofix app is great! Dependabot supports pre-commit hooks, so that’d work! https://github.blog/changelog/2026-03-10-dependabot-now-supports-pre-commit-hooks/
- mypy on pre-commit.ci has the issue of huge env size (limited by pre-commit.ci) and dependency duplication. I found a gh action that syncs the deps, but it’s built with `uv.lock` in mind and its non-locked operating mode is annoying as it pins all dependencies and therefore requires intervention every time. If we find or make a GH action that just syncs the abstract dependencies between pyproject.toml and .pre-commit-config.yaml, that’d work too (with the caveat of that env size limit)

Zethson:
Discussed the overall plan with Phil and we're aligned with what we discussed here. Let's get rid of pre-commit CI. Dependabot supports pre-commit updates which ensures that we still get these updates. The rest will be done with a local mypy run in a GHA workflow as discussed. I'll finish this PR soon.

flying-sheep:
I think we should use the [autofix.ci](https://autofix.ci/) app and not the pre-commit lite one, as its name is more accurate to what it does: both just commit changes to a branch, supporting any possible automatic changes.

{{cookiecutter.project_name}}/.pre-commit-config.yaml
```diff
@@ -23,6 +23,16 @@ repos:
...
       - id: ruff-format
         types_or: [python, pyi, jupyter]
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: uv run mypy src tests
+        language: system
+        types: [python]
+        files: ^(src|tests)/
```
Zethson, line 33: I've also added the tests folder. Not sure what you think of that

{{cookiecutter.project_name}}/pyproject.toml
```diff
@@ -120,6 +121,9 @@ lint.per-file-ignores."docs/*" = [ "I" ]
 lint.per-file-ignores."tests/*" = [ "D" ]
 lint.pydocstyle.convention = "numpy"
 
+[tool.mypy]
+ignore_missing_imports = true
```
Zethson: I think it makes it easier to gradually adopt. Any opinions?

{{cookiecutter.project_name}}/docs/template_usage.md
```diff
@@ -197,15 +197,10 @@ To run the checks locally, we recommend [prek][], a fast, drop-in replacement fo
...
 
-We recommend setting up [pre-commit.ci][] to enforce consistency checks on every commit and pull-request.
-
-To do so, head over to [pre-commit.ci][] and click "Sign In With GitHub".
-Follow the instructions to enable pre-commit.ci for your account or your organization.
-You may choose to enable the service for an entire organization or on a per-repository basis.
-
-Once authorized, pre-commit.ci should automatically be activated.
+The checks also run on every push and pull request through the `Pre-commit checks` job of the `test.yaml` GitHub Actions workflow, so nothing slips through if a contributor forgets to install the git hook.
+This needs no external service or sign-up.
```
flying-sheep: [Claude detected](https://tvtropes.org/pmwiki/pmwiki.php/Main/SuspiciouslySpecificDenial). Please don’t let stuff like this slip through.

Zethson:
Ah sorry @flying-sheep . I hadn't seen that the CI failed. But thanks for fixing it.

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -68,26 +58,17 @@ def elaborate_example(
...
-        if isinstance(item, AnnData):
-            matrix = item.X if not layer_key else item.layers[layer_key]
-        elif isinstance(item, MuData):
-            matrix = item.mod[mudata_mod].X if not layer_key else item.mod[mudata_mod].layers[layer_key]
-        elif isinstance(item, SpatialData):
-            matrix = item.tables[sdata_table_key].X if not layer_key else item.tables[sdata_table_key].layers[layer_key]
-        else:
-            msg = f"Item {item} must be of type AnnData, MuData, or SpatialData but is {item.__class__}."
-            raise ValueError(msg)
+        matrix = item.X if not layer_key else item.layers[layer_key]
```
flying-sheep: with AnnData 0.13:

```suggestion
        matrix = item.layers[layer_key]
```

or even better: replace the parameter with an `LayerAcc | MultiAcc | GraphAcc` and that way support any of these arrays with less code

```py
matrix = adata[acc]
```

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -27,13 +21,11 @@ def basic_preproc(adata: AnnData) -> int:
 
 
 def elaborate_example(
-    items: Iterable[ScverseDataStructures],
-    transform: Callable[[Any], str],
+    items: Iterable[AnnData],
+    transform: Callable[[np.ndarray], str],
     *,  # functions after the asterix are keyword-only arguments
```
flying-sheep: It’s not your change but I just spotted this

Image

{{cookiecutter.project_name}}/pyproject.toml
flying-sheep: yeah, as we discussed, this makes it mostly useless since it basically sneakily infects everything with Any.

`type Foo = A | B | mod.D` and `mod` is ignored? The whole type is now `Any` and will infect everything with `Any`.

You can just use a regular `type:ignore` comment to ignore individual imports, which contains this issue a little bit and is almost as convenient as completely gutting typing with this option, so I’d say get rid of the option.

Zethson: All right then let's remove it. I'll update the PR as soon as I find the time

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
Zethson: I'm okay with a min of AnnData 0.13.0. IDK whether I want to show an accessor here. But no strong opinion.

flying-sheep (approved):

{{cookiecutter.project_name}}/src/{{cookiecutter.package_name}}/pp/basic.py
```diff
@@ -27,13 +21,11 @@ def basic_preproc(adata: AnnData) -> int:
 
 
 def elaborate_example(
-    items: Iterable[ScverseDataStructures],
-    transform: Callable[[Any], str],
-    *,  # functions after the asterix are keyword-only arguments
+    items: Iterable[AnnData],
+    transform: Callable[[np.ndarray], str],
+    *,  # arguments after the asterisk are keyword-only arguments
```
flying-sheep:
```suggestion
    *,  # arguments after the asterisk are keyword-only
```

PR #540, opened 2026-07-24
fix: lint through hatch

### LINKED ISSUE #539: add linting as hatch command
In #519, we moved away from pre-commit CI towards a custom lint job. The CI runs

https://github.com/scverse/cookiecutter-scverse/blob/31733e798d89018e590c2586962c153d38057dad/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/test.yaml#L101

Should we include this into pyproject.toml and switch this to `hatch run lint` to have a single point of truth for these commands?

flying-sheep:
Fixes #539


.github/workflows/test.yml
```diff
@@ -53,7 +52,9 @@ jobs:
         run: |
           cd "$PROJECT_ROOT"
           git add .
-          prek run --verbose --color=always --show-diff-on-failure --all-files
+          hatch check fmt --fix -- --verbose --color=always --show-diff-on-failure
+          hatch check code --fix -- --verbose --color=always --show-diff-on-failure
+          hatch check types -- --color=always
```
flying-sheep: this should probably be reverted except for the “actually typecheck in CI” part

flying-sheep:
@grst @Zethson please take a look

I missed in #530 that this will create a `.venv` in the repo for type checking – nasty! that’ll confuse users to no end!

Zethson (approved):
Thanks for the `.venv` catch!

I love the added groups & the clarification around where to add type stubs. We could already add the pandas one maybe as an example?

PR #553, opened 2026-08-19
add uv.lock to .gitignore

ilia-kats:
The new mypy pre-commit hook uses uv run, which creates a uv.lock file in the project root. That should not end up in Git.


grst (approved):

grst:
btw, feel free to push branches directly to the repo in scverse, then the PR sync action should work, too (in case you didn't have the permission before, it should be fixed now).

PR #130, opened 2022-12-09
improve developers docs

giovp:
this PR does:
- splits the `developers_docs.md` in:
  - `developers_docs.md`  titled __Using this template__ which explains how to go from running `cruft ...` to making having a functional github repository.
  - `contributing.md` titled __Contributing guide__ which explains how to contribute to the repository (from docs building to publishing)
- extend instructions on codecov and add example on manually publishing a python package
 
this PR does not:
- add high level tutorials on python packaging (to be done in separate PR).
- documentation on github actions (to be done in separate PR).
- adding the automated publishing discussed in #88 

actually close #119

PR #558, opened 2026-09-03
feat: Git hooks with autofix.ci & dependabot

flying-sheep:
Switch from pre-commit.ci to dependabot & autofix.ci

also fix a broken link that existed in our post-gen output for 4 years since #130!!!


.github/workflows/propagate-pre-commit.yml
flying-sheep, line 1: should no longer be necessary since dependabot can target both files at once

{{cookiecutter.project_name}}/docs/template_usage.md
```diff
@@ -188,14 +191,24 @@ the project will be published only once you release your package for the first t
...
 
-To run the checks locally, we recommend [prek][], a fast, drop-in replacement for `pre-commit` that reads the same `.pre-commit-config.yaml`.
 Set up the git hook once by running `prek install` (e.g. `uvx prek install`), then `hatch check --fix` to format and check the whole repository.
 
+(autofix-ci)=
+
+### Set up autofix.ci
+
+[autofix.ci][] runs the same checks on every pull request and pushes the automatic fixes back to the pull request branch.
+To enable it, install the [autofix.ci app][] for your repository and follow the instructions there.
```
Zethson, line 207: What happens if the app is not installed but the workflow runs? Nothing or does it crash?

Zethson (approved):
Thank you!

{{cookiecutter.project_name}}/docs/template_usage.md
flying-sheep: I’d assume the `HTTP POST` in the action fails and returns some error message: https://github.com/autofix-ci/action/blob/main/index.ts#L114-L126

PR #561, opened 2026-09-17
build: exclude docs and tests from sdist

ilia-kats:
These are not needed to install the package and can needlessly blow up the size of the source distribution, particularly if docs have multiple tutorial notebooks or the tests include some data files or reference plots.


grst (approved):
Makes sense, thank you!

